PR #43, opened 2026-04-09
feat: enable batch for instrumentation

jeanscherf:
## Description

This PR changes the default to disable_batch=False (async BatchSpanProcessor) and exposes the parameter so users can override it when needed.

auto_instrument was hardcoding disable_batch=True, forcing SimpleSpanProcessor which exports spans synchronously on the request thread. With concurrent workloads, the network round-trip to the OTEL collector adds up directly on the hot path. A user reported response times dropping from ~40s to ~5s under 9 concurrent requests after patching this locally.

## Related Issue

Closes [#40](https://github.com/SAP/cloud-sdk-python/issues/40)

### Type of Change

- Bug fix (non-breaking change that fixes an issue)
- New feature (non-breaking change that adds functionality)
- Documentation update

### How to Test

1. Call auto_instrument() and send concurrent requests to an instrumented agent exporting to an OTLP endpoint
2. Verify response times are not impacted by span export latency
3. Verify auto_instrument(disable_batch=True) still works for cases where synchronous export is needed (e.g. short-lived scripts)

## Checklist

- I have read the Contributing Guidelines
- I have verified that my changes solve the issue
- I have added/updated automated tests to cover my changes
- All tests pass locally
- I have verified that my code follows the Code Guidelines
- I have updated documentation (if applicable)
- I have added type hints for all public APIs
- My code does not contain sensitive information (credentials, tokens, etc.)
- I have followed Conventional Commits for commit messages

## Breaking Changes

None. The new default (disable_batch=False) is a behavior improvement. Users who relied on the previous synchronous behavior can restore it with auto_instrument(disable_batch=True).

## Additional Notes

Trade-offs of BatchSpanProcessor (the new default):
- Spans in the queue are lost if the process is killed abruptly (SIGKILL)
- In short-lived scripts or serverless runtimes, call tracer_provider.force_flush() on shutdown to avoid losing the last batch
- Export errors surface in a background thread (logged, not propagated)

For these cases, disable_batch=True remains available and is documented in the user guide.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "sap-cloud-sdk"
-version = "0.6.0"
+version = "0.6.1"
 description = "SAP Cloud SDK for Python"
 readme = "README.md"
 license = "Apache-2.0"
--- a/src/sap_cloud_sdk/core/telemetry/auto_instrument.py
+++ b/src/sap_cloud_sdk/core/telemetry/auto_instrument.py
@@ -30,12 +30,17 @@ logger = logging.getLogger(__name__)
 
 
 @record_metrics(Module.AICORE, Operation.AICORE_AUTO_INSTRUMENT)
-def auto_instrument():
+def auto_instrument(disable_batch: bool = False):
     """
     Initialize meta-instrumentation for GenAI tracing. Should be initialized before any AI frameworks.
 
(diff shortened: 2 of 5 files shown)


ArthurTonial (approved):

LucasAlvso (approved):

LucasAlvesSoares (approved):
img

PR #71, opened 2026-04-23
feat: propagate custom attributes through non-sdk traces

LucasAlvesSoares:
## Summary

Adds `PropagatedAttributesSpanProcessor` — an OpenTelemetry `SpanProcessor` that injects ContextVar-propagated attributes into every framework-generated child span at `on_start` time.

**Problem**: When `invoke_agent_span(propagate=True)` sets custom attributes (`user.id`, `session.id`, etc.), those attrs are visible on the parent SDK span but _not_ on child spans produced by Traceloop auto-instrumentation (LangChain, Anthropic SDK, etc.) because those frameworks never call `get_propagated_attributes()`.

**Solution**: Register a span processor that reads `_propagated_attrs_var` at the moment each span starts and injects any propagated attributes that are not already present on the span.

**Priority semantics**: propagated attributes are lowest priority — if the framework already set an attribute at span creation time (e.g. `gen_ai.operation.name`, `gen_ai.provider.name`), the child's own value wins. The processor only fills in _missing_ keys.

## Changes

- `src/sap_cloud_sdk/core/telemetry/propagated_attributes_processor.py` — new `PropagatedAttributesSpanProcessor` class
- `src/sap_cloud_sdk/core/telemetry/auto_instrument.py` — registers the processor alongside `BaggageSpanProcessor` in `_set_baggage_processor()`
- `tests/core/unit/telemetry/test_propagated_attributes_processor.py` — unit tests covering injection, skip-if-already-present, non-recording span guard, exception swallowing, and cross-context isolation
- `tests/core/unit/telemetry/test_auto_instrument.py` — added `test_auto_instrument_registers_propagated_attributes_processor`

## Verified end-to-end

Ran a local test (LangChain `ChatLiteLLM` → AI Core, traces exported to Jaeger) and confirmed:
- Parent span (`invoke_agent_span`) carries `user.id`, `session.id`, `test.run`
- Child `ChatLiteLLM.chat` span also carries those same attributes — injected by the processor
- Child's own framework attrs (`gen_ai.operation.name: chat`, `gen_ai.provider.name: litellm`) are _not_ overwritten by propagated values

### DIFF
--- a/.github/workflows/check-version-bump.yaml
+++ b/.github/workflows/check-version-bump.yaml
@@ -23,15 +23,21 @@ jobs:
           CHANGED_FILES=$(git diff --name-only "$BASE_SHA" "$HEAD_SHA")
 
           if echo "$CHANGED_FILES" | grep -q "^src/"; then
-            VERSION_BUMPED=$(git diff -U0 "$BASE_SHA" "$HEAD_SHA" -- pyproject.toml \
-              | grep "^+version = ")
+            BASE_VERSION=$(git show "$BASE_SHA:pyproject.toml" | grep "^version = " | cut -d'"' -f2)
+            HEAD_VERSION=$(git show "$HEAD_SHA:pyproject.toml" | grep "^version = " | cut -d'"' -f2)
 
-            if [ -z "$VERSION_BUMPED" ]; then
-              echo "ERROR: Files under src/ were modified but the version in pyproject.toml was not bumped."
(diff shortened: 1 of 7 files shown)


NicoleMGomes (approved):

NicoleMGomes:
Missing bump version

ArthurTonial (approved):
LGTM, but is missing the version bump

PR #79, opened 2026-04-24
chore: add CI check to enforce version bump when src/ is modified

jplbrun:
## Description

Added a GitHub Actions workflow that fails the PR if any file under src/ was changed but the version = "..." line in pyproject.toml was not modified

### DIFF
--- a/.github/workflows/check-version-bump.yaml
+++ b/.github/workflows/check-version-bump.yaml
@@ -32,4 +32,6 @@ jobs:
             else
               echo "Version bump detected: $VERSION_BUMPED"
             fi
+          else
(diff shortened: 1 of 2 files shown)


jeanscherf (approved):
Great work on this, @jplbrun  Very clean implementation. This is exactly what we needed.

Two small things I wanted to comment:

- When src/ is not modified, the job exits silently with no output. 
Maybe an extra else with `echo "No src/ changes. Version bump not required."`  would make the CI logs self-explanatory.
- The file is named .yml while all other workflows in this repo use .yaml. Maybe it's worth aligning for consistency.

Neither is a blocker. Thanks for picking this up!

PR #90, opened 2026-05-05
feat(secret-resolver): support SERVICE_BINDING_ROOT env var for binding discovery

cassiofariasmachado:

## Description

Adds support for the [`SERVICE_BINDING_ROOT`](https://servicebinding.io/spec/core/1.1.0/) environment variable as a configurable base path for service binding discovery, aligning the Python Cloud SDK with the servicebinding.io spec and other SAP SDKs (Java).

- Introduces a public `resolve_base_mount(default)` helper in `resolver.py` that returns `SERVICE_BINDING_ROOT` if set, otherwise falls back to the provided default (`/etc/secrets/appfnd`)
- Wires `resolve_base_mount` into `read_from_mount_and_fallback_to_env_var()` so all callers (destination, objectstore, auditlog, dms) benefit automatically without any changes at call sites
- Exports `resolve_base_mount` from `sap_cloud_sdk.core.secret_resolver` so other modules can reuse the same logic
- Updates `aicore/__init__.py` to import and use `resolve_base_mount` instead of a hardcoded path
- Documents the `SERVICE_BINDING_ROOT` env var in the secret resolver user guide
- Adds two unit tests covering the override and fallback behaviour

## Related Issue

N/A

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Set `SERVICE_BINDING_ROOT=/some/custom/path` in the environment
2. Call any service that uses the secret resolver (e.g. objectstore, destination)
3. Verify the resolver looks for secrets under `/some/custom/path/<module>/<instance>/` instead of `/etc/secrets/appfnd/`
4. Unset `SERVICE_BINDING_ROOT` and verify it falls back to `/etc/secrets/appfnd`
5. Run unit tests: `uv run pytest tests/core/unit/secret_resolver/ tests/aicore/ -m "not integration" -q`

## Checklist

- [x] All tests pass locally

## Breaking Changes

None. `SERVICE_BINDING_ROOT` is opt-in — when not set, behaviour is identical to before.

## Additional Notes

The `resolve_base_mount` function is exported publicly so that modules with custom secret resolution logic (like `aicore`) can reuse the same `SERVICE_BINDING_ROOT` lookup without duplicating the env var read.

### DIFF
--- a/src/sap_cloud_sdk/core/secret_resolver/resolver.py
+++ b/src/sap_cloud_sdk/core/secret_resolver/resolver.py
@@ -7,6 +7,11 @@ from dataclasses import fields, is_dataclass
 from typing import Any, Dict, Tuple
 
 
+def resolve_base_mount(default: str) -> str:
+    """Return SERVICE_BINDING_ROOT if set, otherwise the provided default."""
+    return os.environ.get("SERVICE_BINDING_ROOT", default)
+
+
 def _validate_inputs(module: str, instance: str) -> None:
     """Validate module and instance inputs."""
     if not isinstance(module, str) or not module.strip():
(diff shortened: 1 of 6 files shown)


src/sap_cloud_sdk/core/secret_resolver/resolver.py
```diff
@@ -7,6 +7,11 @@
 from typing import Any, Dict, Tuple
 
 
+def resolve_base_mount(default: str) -> str:
```
NicoleMGomes, line 10: I believe default path could be created as a constant here, instead of a parameter.

PR #109, opened 2026-05-12
fix: adapt event_type to be fetched from descriptop

simeongelovski:
Updated the auditlog-ng client audit log event name handling to derive the event name directly from the protobuf event descriptor.

Previously, event names were in the format:

`sap.als.AuditEvent.<EVENT_TYPE>.v2
`
With this change, event names now follow the protobuf descriptor naming convention:

`sap.auditlog.auditevent.v2.<EVENT_TYPE>
`
Example:

`sap.auditlog.auditevent.v2.ConfigurationChange
`
This aligns the OTEL event name format with the ALS protobuf schema and the Audit Log SDK implementation, where the event type is resolved from the protobuf DESCRIPTOR.

#### Breaking Changes

The auditlog-ng client no longer supports user-provided audit event types. Event names are now derived exclusively from the protobuf event descriptor (DESCRIPTOR) and follow the protobuf schema naming convention:


src/sap_cloud_sdk/core/auditlog_ng/client.py
```diff
@@ -225,9 +219,9 @@ def send(
 
         return event_id
 
-    def send_json(self, event: Message, event_type: Optional[str] = None) -> str:
+    def send_json(self, event: Message) -> str:
         """Send event in JSON format."""
-        return self.send(event, event_type, format="json")
+        return self.send(event, format="json")
```
jeanscherf, line 224: This is a breaking change, right? Shouldn't we document that?

src/sap_cloud_sdk/core/auditlog_ng/client.py
betinacosta: this line should be removed, right?

src/sap_cloud_sdk/core/auditlog_ng/client.py
simeongelovski: done

betinacosta (dismissed):

betinacosta (approved):

PR #133, opened 2026-05-21
feat: add configurable token cache for customer and lob flows

vitalykumov:
## Description

Add in-process token cache for both customer and LoB agent flows. Previously every `list_mcp_tools()` / `call_mcp_tool()` call fetched fresh IAS token via mTLS - unnecessary latency in agentic loops.

**Changes:**

- **`_token_cache.py`** (new): `_TokenCache` - TTL + LRU eviction for system tokens (key: `client_id`) and user tokens (key: `sha256(user_jwt+"|"+client_id)[:16]`). Expiry is from `expires_in`, `id_token` exp claim, or fallback TTL.
- **`_customer.py`**: `get_system_token_mtls` / `exchange_user_token` to consult/populate cache. 401 response from MCP server → invalidate stale token + retry once
- **`_lob.py`**: `get_system_auth` / `get_user_auth` now check cache before calling BTP Destination Service. `get_mcp_tools_lob` reuses system auth across all fragments (single fetch per discovery call). `call_mcp_tool_lob` caches user token per `(fragment, tenant)` scope. Both functions retry once on 401 with cache invalidation - same pattern as customer flow
- **`agw_client.py`**: `AgentGatewayClient` passes `_TokenCache` to both customer and LoB functions
- **`config.py`**: 4 new `ClientConfig` fields added - `token_expiry_buffer_seconds` (60 s), `max_system_token_cache_size` (10 s), `max_user_token_cache_size` (10 s), `fallback_token_ttl_seconds` (300 s)

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

Describe how reviewers can test your changes:

1. Programmatically:
    ```python
    async def verify_token_caching(agw_client) -> bool:
        """Verify system token is cached across list_mcp_tools calls.
    
        Patches _request_token_mtls to count real HTTP token requests.
        First call must fetch at least one token; second call must reuse cache.
        Returns True if caching works correctly.
        """
        original = _customer_mod._request_token_mtls
        call_count = 0
    
        def counting(*args, **kwargs):
            nonlocal call_count
            call_count += 1
            return original(*args, **kwargs)
    
        _customer_mod._request_token_mtls = counting
        try:
            await agw_client.list_mcp_tools()
            after_first = call_count
    
            await agw_client.list_mcp_tools()
            after_second = call_count
    
            agw_client.clear_token_cache()
            await agw_client.list_mcp_tools()
            after_clear = call_count
        finally:
            _customer_mod._request_token_mtls = original
    
        cache_hit = after_second == after_first
        invalidate_works = after_clear > after_second
    
        print("\nToken Caching")
        print("-" * 40)
        print(f"  1st call  token requests : {after_first}")
        print(f"  2nd call  token requests : {after_second - after_first}  {'✓ cache hit' if cache_hit else '✗ expected 0'}")
        print(f"  post-clear token requests: {after_clear - after_second}  {'✓ re-fetched' if invalidate_works else '✗ expected ≥1'}")
    
        return cache_hit and invalidate_works
    ...
    async def main():
        agw_client = create_client()
    
        ok = await verify_token_caching(agw_client)
        if not ok:
            print("\nWARNING: token caching check failed")
    ...
    ```
2. Unit tests:
   ```bash
   pytest tests/agentgateway/unit/test_token_cache.py tests/agentgateway/unit/test_customer.py tests/agentgateway/unit/test_agw_client.py
   ```

## Checklist

- [x] All tests pass locally

## Breaking Changes

None. Cache internal to `AgentGatewayClient` - existing `create_client()` calls get caching automatically. `ClientConfig` new fields all have defaults.

## Additional Notes

**Thread safety:** GIL makes individual `OrderedDict` ops atomic, but check-then-set is not. Concurrent coroutines on same key may both miss and both fetch - redundant requests, not corruption. Acceptable for agentic loop use case.

**401 retry:** Both `get_mcp_tools_customer` and `call_mcp_tool_customer` invalidate + retry once on 401, handling server-side revocation before token expiry.

### DIFF
--- a/src/sap_cloud_sdk/agentgateway/_customer.py
+++ b/src/sap_cloud_sdk/agentgateway/_customer.py
@@ -320,9 +320,9 @@ def get_system_token_mtls(
     Returns:
         System-scoped access token.
     """
-    cached = cache.get_system_token(app_tid)
+    cached = cache.get_system_token(credentials.client_id)
     if cached:
-        logger.debug("Using cached system token (app_tid=%s)", app_tid)
+        logger.debug("Using cached system token (client_id=%s)", credentials.client_id)
         return cached
 
     logger.info("Acquiring system token via mTLS client credentials")
@@ -334,7 +334,7 @@ def get_system_token_mtls(
         app_tid=app_tid,
         extra_data={"response_type": "token"},
     )
-    cache.set_system_token(token, expires_at, app_tid)
+    cache.set_system_token(token, expires_at, credentials.client_id)
     return token
 
 
@@ -363,9 +363,9 @@ def exchange_user_token(
     Returns:
         AGW-scoped access token with user identity.
     """
-    cached = cache.get_user_token(user_token, app_tid)
+    cached = cache.get_user_token(user_token, credentials.client_id)
     if cached:
-        logger.debug("Using cached user token (app_tid=%s)", app_tid)
+        logger.debug("Using cached user token (client_id=%s)", credentials.client_id)
         return cached
 
     logger.info("Exchanging user token for AGW-scoped token via jwt-bearer grant")
@@ -380,7 +380,7 @@ def exchange_user_token(
             "token_format": "jwt",
         },
     )
-    cache.set_user_token(user_token, token, expires_at, app_tid)
+    cache.set_user_token(user_token, token, expires_at, credentials.client_id)
     return token
 
 
@@ -520,7 +520,7 @@ async def get_mcp_tools_customer(
             dep.global_tenant_id,
         )
 
-        while True:
+        for attempt in (1, 2):
             if not system_token:
                 # won't catch exceptions here - if token acquisition fails,
                 # we want the discovery to fail immediately
@@ -532,12 +532,12 @@ async def get_mcp_tools_customer(
                 logger.debug("Loaded %d tool(s) from %s", len(server_tools), dep.ord_id)
             except Exception as exc:
                 unwrapped = _unwrap_exception_group(exc)
-                if _is_unauthorized(unwrapped):
+                if _is_unauthorized(unwrapped) and attempt == 1:
                     logger.info(
                         "401 from %s — invalidating cached system token and retrying",
                         dep.ord_id,
                     )
-                    cache.invalidate_system_token(app_tid)
+                    cache.invalidate_system_token(credentials.client_id)
                     system_token = None  # Force refetch on next loop iteration
                     continue
                 logger.exception("Failed to load tools from %s — skipping", dep.ord_id)
@@ -609,9 +609,9 @@ async def call_mcp_tool_customer(
 
     def _invalidate_token() -> None:
         if user_token:
-            cache.invalidate_user_token(user_token, app_tid)
+            cache.invalidate_user_token(user_token, credentials.client_id)
         else:
-            cache.invalidate_system_token(app_tid)
+            cache.invalidate_system_token(credentials.client_id)
 
     last_exc: Exception | None = None
     for attempt in (1, 2):
--- a/src/sap_cloud_sdk/agentgateway/_customer.py
+++ b/src/sap_cloud_sdk/agentgateway/_customer.py
@@ -499,11 +499,15 @@ async def get_mcp_tools_customer(
 
     logger.info("Discovering tools from %d MCP server(s)", len(dependencies))
 
-    # Get system token for discovery
-    loop = asyncio.get_running_loop()
-    system_token = await loop.run_in_executor(
-        None, get_system_token_mtls, credentials, timeout, config, cache, app_tid
-    )
+    system_token = None
+    # Define a helper closure to refetch the system token on demand, since it may need to be
+    # refreshed during the discovery loop if any server returns a 401
+    async def refetch_system_token() -> str:
+        loop = asyncio.get_running_loop()
+        new_token = await loop.run_in_executor(
+            None, get_system_token_mtls, credentials, timeout, config, cache, app_tid
+        )
+        return new_token
 
     tools: list[MCPTool] = []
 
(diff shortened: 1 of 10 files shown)


src/sap_cloud_sdk/agentgateway/_customer.py
```diff
@@ -292,61 +301,87 @@ def _request_token_mtls(
...
         credentials: Customer credentials.
         timeout: HTTP timeout in seconds.
+        config: Client configuration.
+        cache: Token cache to consult and update.
         app_tid: BTP Application Tenant ID of subscriber (optional).
 
     Returns:
         System-scoped access token.
     """
+    cached = cache.get_system_token(app_tid)
```
prashantrakheja: i am not sure if app_tid would be needed for fetching the tokens, is there any other way to do this?

src/sap_cloud_sdk/agentgateway/_customer.py
```diff
@@ -480,7 +520,42 @@ async def get_mcp_tools_customer(
             server_tools = await _list_server_tools(url, system_token, dep, timeout)
             tools.extend(server_tools)
             logger.debug("Loaded %d tool(s) from %s", len(server_tools), dep.ord_id)
-        except Exception:
+        except Exception as exc:
+            unwrapped = _unwrap_exception_group(exc)
```
prashantrakheja: do u feel the exception block is unusually big? can we put this stuff inside a method?

src/sap_cloud_sdk/agentgateway/_customer.py
```diff
@@ -480,7 +520,42 @@ async def get_mcp_tools_customer(
...
+                    # Replace stale token for remaining iterations
+                    system_token = fresh_token
+                    logger.debug(
+                        "Loaded %d tool(s) from %s after retry",
+                        len(server_tools),
+                        dep.ord_id,
+                    )
+                    continue
+                except Exception:
+                    logger.exception(
```
prashantrakheja: i feel we should not swallow the exception here, if there is a failure we should fail-fast, what's ur take?

src/sap_cloud_sdk/agentgateway/agw_client.py
```diff
@@ -85,6 +86,17 @@ def __init__(
         """
         self._tenant_subdomain = tenant_subdomain
         self._config = config or ClientConfig()
+        self._token_cache = _TokenCache(self._config)
+
+    @record_metrics(Module.AGENTGATEWAY, Operation.AGENTGATEWAY_CLEAR_TOKEN_CACHE)
```
prashantrakheja: this operation is "internal" to the functioning of the sdk, do u feel it makes sense to capture metrics for this?

prashantrakheja (commented):
changes look good, some minor feedback, in particular i would like to remove usage of app_tid from token caching because its an optional parameter and likely to be removed going forward..

src/sap_cloud_sdk/agentgateway/agw_client.py
cassiofariasmachado: I didn't see value exposing this method

cassiofariasmachado (changes requested):

src/sap_cloud_sdk/agentgateway/agw_client.py
vitalykumov: Removed. My assumption was that consumers of SDK should be able to set/unset/cleanup cache should they require it. I had an impression that it's usual for SDKs :) removed as considered unwanted

src/sap_cloud_sdk/agentgateway/_customer.py
vitalykumov: I'll extract logic to a method, good catch.
Regarding exception - it basically follows what was implemented before - log with `logger.exception` and continue

src/sap_cloud_sdk/agentgateway/_customer.py
vitalykumov: Updated logic here - request system_token when needed. Added helper closure to refetch token

src/sap_cloud_sdk/agentgateway/_customer.py
vitalykumov: switched to use `client_id`

src/sap_cloud_sdk/agentgateway/_customer.py
```diff
@@ -211,19 +213,24 @@ def _request_token_mtls(
     credentials: CustomerCredentials,
     grant_type: str,
     timeout: float,
+    config: ClientConfig,
```
NicoleMGomes: Isn't timeout part of client config?

src/sap_cloud_sdk/agentgateway/_customer.py
```diff
@@ -25,6 +25,8 @@
     IntegrationDependency,
```
NicoleMGomes, line 25: Shouldn't we also add token cache for LoB flow to keep consistency?

src/sap_cloud_sdk/agentgateway/_customer.py
vitalykumov: Yes, though it's not needed here. Passing cache instead now since it's available and has required information

src/sap_cloud_sdk/agentgateway/_customer.py
vitalykumov: I had an impression that since LoB flow goes through destination service it would handle the token request and caching. However seems like it's not the case. I'll add caching to lob flow

src/sap_cloud_sdk/agentgateway/_customer.py
vitalykumov: @NicoleMGomes I added caching to LoB flow as well :)

prashantrakheja (approved):

NicoleMGomes (approved):

cassiofariasmachado (approved):

cassiofariasmachado:
@vitalykumov, could you take a look at the conflicts?

vitalykumov:
> @vitalykumov, could you take a look at the conflicts?

I'll try but seems like it would require to change quite a lot...

Quick look:
Token fetching is separate now, doesn't seem like there's any reason to keep caching in customer/lob flows. I'll need to re-write pretty much everything. I'll continue on Monday

vitalykumov:
closing in favor of https://github.com/SAP/cloud-sdk-python/pull/142

PR #137, opened 2026-05-25
feat: use dedicated fragments for AGW token generation

NicoleMGomes:
PR recreated from contrib #110 extended with tests

## Description

Adding token fetching based on dedicated AGW fragments for the AGW client. Additionally it exposes the URL of AGW as fetched from the fragment.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Use an agent that has the new fragment generation applied
2. Make sure that these fragments were created for your test tenant
3. Do a regression test of the existing client functionality
4. Also test whether the standalone auth fetching functions work and return valid tokens + URLs

## Checklist

- [x] All tests pass locally

### DIFF
--- a/src/sap_cloud_sdk/agentgateway/_lob.py
+++ b/src/sap_cloud_sdk/agentgateway/_lob.py
@@ -392,11 +392,6 @@ async def call_mcp_tool_lob(
     Returns:
         Tool execution result as string.
     """
-    if not tool.fragment_name:
-        raise MCPServerNotFoundError(
-            f"Tool '{tool.name}' missing fragment_name for LoB invocation"
-        )
-
     async with httpx.AsyncClient(
         headers={
             "Authorization": f"Bearer {user_auth_token}",
--- a/src/sap_cloud_sdk/agentgateway/_lob.py
+++ b/src/sap_cloud_sdk/agentgateway/_lob.py
@@ -6,7 +6,6 @@ LoB agents use BTP Destination Service for credential management:
 """
 
 import asyncio
-import base64
 import logging
 import os
 import uuid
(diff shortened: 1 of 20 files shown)


cassiofariasmachado (dismissed):

src/sap_cloud_sdk/agentgateway/_lob.py
```diff
@@ -328,36 +375,33 @@ async def get_mcp_tools_lob(
...
+        timeout: HTTP timeout in seconds for the MCP server call.
         **kwargs: Tool input parameters.
 
     Returns:
         Tool execution result as string.
-
-    Raises:
-        MCPServerNotFoundError: If destination/auth fails.
     """
     if not tool.fragment_name:
```
prashantrakheja: since u removed the line below where fragment_name was actually required, is this check still relevant?

tests/agentgateway/integration/agw_auth.feature
```diff
@@ -0,0 +1,54 @@
...
+    When I call list_mcp_tools
+    Then the result should be a list of MCPTool
+    And the list should be non-empty
+    And each tool should have a non-empty name
+    And each tool should have a non-empty url
+
+  Scenario: Call search_workflows tool returns a non-empty result
+    Given I have a valid user token
+    When I call list_mcp_tools
+    And I call call_mcp_tool with "search_workflows" and the user token
```
prashantrakheja, line 53: this test should be removed or updated, it relies on a very specific tool which may not be available in all mcp servers

prashantrakheja (commented):
looks good, tested customer flow with system credentials, minor comments

tests/agentgateway/integration/agw_auth.feature
NicoleMGomes: This is not a problem, because the credentials are specific env vars we set and it today will use the a2a-agent consuming N8N. In case it changes in future, then we will update.

src/sap_cloud_sdk/agentgateway/_lob.py
NicoleMGomes: Removed

prashantrakheja (approved):

cassiofariasmachado (approved):

PR #142, opened 2026-06-01
feat(agentgateway): add client-owned token cache

vitalykumov:
## Description

Adds a per-client token cache to `AgentGatewayClient` to avoid redundant IAS token requests during agentic loops. Both the customer (mTLS) and LoB (BTP Destination Service) flows are covered.

**Token cache (`_token_cache.py`):**
- `_TokenCache` - LRU-bounded `OrderedDict` for system and user tokens with monotonic TTL. Expiry resolves from `expires_at` → `expires_in` → `exp` JWT claim (access then id token) → config fallback TTL. Configurable buffer subtracted from all explicit expiries.
- `_GatewayUrlCache` - separate LRU-bounded `OrderedDict` for gateway URLs (stable within a client's lifetime, bounded to prevent memory growth in long-lived multi-tenant clients).
- `_parse_jwt_exp` - unverified JWT payload parser used only for TTL hints, never for security decisions.

**Customer flow (`_customer.py`):**
- `_request_token_mtls` now returns the full token response dict instead of just `access_token`, so callers can pass it to `compute_expires_at`.
- `get_system_token_mtls` and `exchange_user_token` accept an optional `_TokenCache` and short-circuit on cache hit.

**LoB flow (`_lob.py`):**
- `fetch_system_auth` and `fetch_user_auth` accept optional `_TokenCache` + `_GatewayUrlCache` and short-circuit on cache hit.
- Both functions raise `ValueError` if exactly one of `token_cache`/`gateway_url_cache` is provided (previously the cache was silently skipped in that case).

**Client (`agw_client.py`):**
- `AgentGatewayClient.__init__` creates a `_TokenCache` and `_GatewayUrlCache` instance and threads them through all auth calls.

**Config (`config.py`):**
- Four new `ClientConfig` fields: `fallback_token_ttl_seconds` (300 s), `token_expiry_buffer_seconds` (30 s), `max_system_token_cache_size` (32), `max_user_token_cache_size` (256).

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Run the unit test suite: `uv run pytest tests/agentgateway/unit/ -q`
3. To confirm cache hit behaviour manually: instantiate `AgentGatewayClient`, call `get_system_auth` twice with the same tenant/credentials, observe only one token request is made (mock or inspect the `_token_cache` directly via `client._token_cache`).

## Checklist

- [x] All tests pass locally

## Breaking Changes

None. All cache params are optional with `None` defaults. Existing callers unaffected. `_request_token_mtls` return type changed from `str` to `dict`, but it is a private function.

## Additional Notes

- User token cache keys are `sha256(user_jwt + "|" + scope_key)[:16]`.
- Thread safety: CPython GIL makes individual `OrderedDict` ops atomic, but check-then-set is not. Concurrent coroutines for the same key may both miss and both fetch - produces redundant requests, not corruption. Documented in module docstring.
- `_GatewayUrlCache` default max size is 64 (independent of `ClientConfig`). Can be made configurable if needed.

### DIFF
--- a/src/sap_cloud_sdk/agentgateway/_token_cache.py
+++ b/src/sap_cloud_sdk/agentgateway/_token_cache.py
@@ -92,7 +92,8 @@ def _parse_response_expires_at(expires_at: object) -> float | None:
     return parsed.timestamp()
 
 
-def _monotonic_expiry_from_epoch(expiry_epoch_seconds: float, buffer: float) -> float:
+def _monotonic_expiry_from_epoch(expiry_epoch_seconds: float,
+                                 buffer: float) -> float:
     """Translate a wall-clock expiry into a monotonic deadline."""
     return time.monotonic() + (expiry_epoch_seconds - time.time()) - buffer
 
@@ -160,9 +161,6 @@ class _GatewayUrlCache:
         while len(self._cache) > self._max_size:
             self._cache.popitem(last=False)
 
-    def __contains__(self, scope_key: str) -> bool:
-        return scope_key in self._cache
-
 
 class _TokenCache:
     """Per-client token cache with TTL and LRU eviction.
@@ -188,13 +186,14 @@ class _TokenCache:
             del self._system_tokens[scope_key]
         return None
 
-    def set_system_token(self, token: str, expires_at: float, scope_key: str) -> None:
+    def set_system_token(self, token: str, expires_at: float,
+                         scope_key: str) -> None:
         """Cache a system token under `scope_key`; evict LRU once size exceeds limit."""
-        self._system_tokens[scope_key] = _CachedToken(
-            token=token, expires_at=expires_at
-        )
+        self._system_tokens[scope_key] = _CachedToken(token=token,
+                                                      expires_at=expires_at)
         self._system_tokens.move_to_end(scope_key)
-        while len(self._system_tokens) > self._config.max_system_token_cache_size:
+        while len(self._system_tokens
+                  ) > self._config.max_system_token_cache_size:
             evicted, _ = self._system_tokens.popitem(last=False)
             logger.debug("System token cache full — evicted '%s'", evicted)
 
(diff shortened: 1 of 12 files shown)


cassiofariasmachado (dismissed):

NicoleMGomes (dismissed):

src/sap_cloud_sdk/agentgateway/user-guide.md
```diff
@@ -99,10 +100,38 @@ The SDK automatically detects the agent type based on the presence of a credenti
...
+    max_system_token_cache_size=32,
+    max_user_token_cache_size=256,
+)
+
+agw_client = create_client(tenant_subdomain="my-tenant", config=config)
+```
+
+- `timeout`: HTTP timeout in seconds for token requests and MCP calls. Default: `60.0`.
+- `fallback_token_ttl_seconds`: Used when the token response does not include expiry metadata. Default: `300.0`.
+- `token_expiry_buffer_seconds`: Safety buffer subtracted from explicit token expiries before a cached token is reused. Default: `30.0`.
```
prashantrakheja, line 130: do u think we need to add a validation ensuring token_expiry_buffer_seconds is not greater than fallback_token_ttl_seconds ?

src/sap_cloud_sdk/agentgateway/_token_cache.py
```diff
@@ -0,0 +1,276 @@
...
+            self._cache.move_to_end(scope_key)
+        return value
+
+    def __setitem__(self, scope_key: str, url: str) -> None:
+        self._cache[scope_key] = url
+        self._cache.move_to_end(scope_key)
+        while len(self._cache) > self._max_size:
+            self._cache.popitem(last=False)
+
+    def __contains__(self, scope_key: str) -> bool:
```
prashantrakheja: is this used anywhere in our code?

prashantrakheja (approved):
couple of minor comments, looks good otherwise

src/sap_cloud_sdk/agentgateway/user-guide.md
vitalykumov: added validation

src/sap_cloud_sdk/agentgateway/_token_cache.py
vitalykumov: removed `__contains__`

cassiofariasmachado (approved):

PR #155, opened 2026-06-10
chore: resolve pre-commit hook failures and enforce them in CI

jeanscherf:
## Summary

- Fix all `ty` type check errors that were blocking `pre-commit`:
  - Use `MessageRole.USER` enum instead of the `"USER"` string literal in agent memory BDD test
  - Add `assert token is not None` before `lambda: token` to narrow `Optional[str]` to `str` (matches `Callable[[], str]`)
  - Move top-level `langgraph`/`langchain_core` imports inside the `try/except` block in `_agent.py` and add `# ty: ignore[unresolved-import]` for optional deps not present in the venv
  - Annotate `DummyResponse.headers` as `Mapping[str, str]` so it satisfies the `_ResponseLike` protocol
  - Remove two now-stale `# ty: ignore` suppression comments in DMS integration tests
- Replace the individual `ruff` and `ty` steps in `checks.yaml` with a single `pre-commit run --all-files` step, which also enforces YAML validity and trailing-whitespace checks that were previously not covered in CI

## Test plan

- [x] All pre-commit hooks pass locally (`VIRTUAL_ENV=.venv pre-commit run --all-files`)
- [x] `checks.yaml` CI job passes on this PR


LucasAlvesSoares (dismissed):
img

ArthurTonial (approved):

PR #156, opened 2026-06-11
feat(extensibility): implement call_hook_agw

yashmeet29:

## Description

Adds `call_hook_agw`, an async method on `ExtensibilityClient` that calls an n8n hook via the Agent Gateway (AGW) MCP transport, as an alternative to the existing `call_hook` method which uses a direct HTTP connection.

The method:

1. Creates an AGW client for the given tenant subdomain via `create_agw_client`.
2. Discovers MCP tools from the AGW, using a composite `(name, server_name)` key to match `execute_workflow` and `get_execution` tools on the `sap.btpn8n:apiResource:ManagedN8nMcpServer:v1` server specifically.
3. Invokes `execute_workflow` with the workflow ID, method, body (serialised from the optional Message payload), and headers.
4. After checking for terminal errors in the initial execute_workflow response, always enters the get_execution poll loop.
5. Polls get_execution every 500 ms until the execution reaches success (extracts and returns a Message from the last executed node), a terminal status, or hook.timeout seconds elapse — raising TransportError in all failure/timeout cases.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Extend the Agent using the [Extensibility Skill](https://github.tools.sap/application-foundation/agent-skills/tree/main/.claude/skills/sap-agent-extensibility)
2. Extend the agent with [pre/post hooks](https://pages.github.tools.sap/customforms/agent-extensibility-documentation/use-cases/pre-post-hooks.html)
3. Call the Agent through AgentGateway (A2A)
4. Verify the hooks were executed and the Agent response is correct

## Checklist

- [x] All tests pass locally

## Additional Notes
`call_hook_agw` uses `Operation.EXTENSIBILITY_CALL_HOOK` for telemetry — the same operation name as the existing `call_hook`. This is intentional reuse of a pre-existing enum value; no new telemetry operation was introduced.

Tool lookup uses a composite key `(name, server_name)` to avoid false matches when multiple MCP servers expose tools with the same name. The server name is `sap.btpn8n:apiResource:ManagedN8nMcpServer:v1`.

Two distinct terminal-status sets are used: `_EXECUTE_TERMINAL_STATUSES` (checked against the initial `execute_workflow` response, includes `"unknown"`) and `_EXECUTION_TERMINAL_STATUSES` (checked during the poll loop, excludes `"unknown"`).

11 new unit tests cover: tool-not-found for both tools, composite-key correctness (wrong server ignored, correct tool selected among duplicates), synchronous success, success after polling, terminal status from execute and from poll, timeout, AGW exception propagation, and workflow ID forwarding.


cassiofariasmachado:
> Implement callHook function via Agent Gateway

Please provide a more detailed description of this pull request using our template [pull_request_template.md](https://github.com/SAP/cloud-sdk-python/blob/main/.github/pull_request_template.md)

src/sap_cloud_sdk/extensibility/client.py
```diff
@@ -38,6 +40,7 @@
 
 _EXECUTE_WORKFLOW_TOOL_NAME = "execute_workflow"
 _GET_EXECUTION_TOOL_NAME = "get_execution"
+_N8N_MCP_SERVER_NAME = "sap.btpn8n:apiResource:ManagedN8nMcpServer:v1"
```
NicoleMGomes, line 43: ORD ID is dynamically discovered using SPII. As far as I know, agent should also have formation with n8n.

src/sap_cloud_sdk/extensibility/client.py
```diff
@@ -410,3 +413,216 @@ def call_hook(
...
+        execute_tool = next(
+            (
+                t
+                for t in tools
+                if t.name == _EXECUTE_WORKFLOW_TOOL_NAME
+                and t.server_name == _N8N_MCP_SERVER_NAME
+            ),
+            None,
+        )
+        if execute_tool is None:
```
NicoleMGomes, line 413: Let's breakdown in functions and organize code

NicoleMGomes (changes requested):

src/sap_cloud_sdk/extensibility/client.py
rishikunnath2747: Agent does have a formation with n8n, but other mcp servers may also be part of the formation. In that case we need to use the ORD ID to filter through the fragments.

Also this ORD ID will be constant for hooks, they are only using n8n for now

src/sap_cloud_sdk/extensibility/client.py
yashmeet29: Done

src/sap_cloud_sdk/extensibility/client.py
```diff
@@ -410,3 +395,231 @@ def call_hook(
             f"Workflow execution timed out after {hook.timeout}s. "
             f"Last status: {last_status!r}"
         )
+
+    async def _discover_n8n_tools(
+        self, agw_client: Any, user_token: Optional[str]
+    ) -> tuple[Any, Any]:
+        # TODO: cache the list of mcp tools for performance.
```
NicoleMGomes: Cache will be offered as part of agent gateway module. So I believe you can remove this TODO and we will let you know once it's available.

src/sap_cloud_sdk/extensibility/client.py
```diff
@@ -410,3 +395,231 @@ def call_hook(
             f"Workflow execution timed out after {hook.timeout}s. "
             f"Last status: {last_status!r}"
         )
+
+    async def _discover_n8n_tools(
+        self, agw_client: Any, user_token: Optional[str]
```
NicoleMGomes, line 400:
```suggestion
        self, 
        agw_client: Any, 
        user_token: Optional[str]
```

src/sap_cloud_sdk/extensibility/client.py
```diff
@@ -410,3 +395,231 @@ def call_hook(
...
+            (
+                t
+                for t in tools
+                if t.name == _EXECUTE_WORKFLOW_TOOL_NAME
+                and t.server_name == _N8N_MCP_SERVER_NAME
+            ),
+            None,
+        )
+        if execute_tool is None:
+            raise TransportError(
```
NicoleMGomes: Is this a transport error? Because tool was not found, but request to agw probably was successful

src/sap_cloud_sdk/extensibility/client.py
```diff
@@ -410,3 +395,231 @@ def call_hook(
...
+            (
+                t
+                for t in tools
+                if t.name == _GET_EXECUTION_TOOL_NAME
+                and t.server_name == _N8N_MCP_SERVER_NAME
+            ),
+            None,
+        )
+        if get_exec_tool is None:
+            raise TransportError(
```
NicoleMGomes: Double-check exceptions across all functions to be more accurate for cases

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "sap-cloud-sdk"
-version = "0.29.0"
+version = "0.29.1"
```
NicoleMGomes:
```suggestion
version = "0.30.0"
```
It's not a fix, it's a new functionality

NicoleMGomes:
Review `user-guide.md`

pyproject.toml
rishikunnath2747: Changed

src/sap_cloud_sdk/extensibility/client.py
rishikunnath2747: Changed to ExtensibilityError instead

src/sap_cloud_sdk/extensibility/client.py
rishikunnath2747: Changed to ExtensibilityError instead

src/sap_cloud_sdk/extensibility/client.py
rishikunnath2747: This causes the ruff check to fail

NicoleMGomes: Probably because of empty spaces, because format is valid and used across multiple methods. But this is optional

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "sap-cloud-sdk"
-version = "0.29.1"
+version = "0.30.0"
```
NicoleMGomes:
```suggestion
version = "0.31.0"
```

Sorry, we just released version `0.30.0`

NicoleMGomes (dismissed):

NicoleMGomes (approved):

PR #169, opened 2026-06-16
feat(auditlog_ng): support destination based configuration

betinacosta:
## Description

The AuditLog NG module currently requires callers to explicitly provide endpoint, deployment_id, and namespace when calling create_client. In SPII-based deployments, these values are stored in a Destination (with tenant-specific fragments) following the ALS SPII v3.0.0 provisioning lifecycle. This change allows create_client to resolve these parameters automatically from a Destination, reducing boilerplate and avoiding misconfiguration.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

```python
from sap_cloud_sdk.core.auditlog_ng import create_client

client = create_client(
    destination_name="AuditLogV3_Destination",
    destination_instance="default"
)

print(f"Client created: {client}")
print(f"  endpoint:      {client._config.endpoint}")
print(f"  deployment_id: {client._config.deployment_id}")
print(f"  namespace:     {client._config.namespace}")
```

## Checklist

- [ ] I have verified that my changes solve the issue
- [ ] All tests pass locally
- [ ] I have verified that my code follows the [Code Guidelines](../docs/GUIDELINES.md)
- [ ] My code does not contain sensitive information (credentials, tokens, etc.)
- [ ] I have followed [Conventional Commits](https://www.conventionalcommits.org/) for commit messages

## Breaking Changes

If this PR introduces breaking changes, please describe:

- What breaks
- Migration path for users
- Alternative approaches considered

## Additional Notes

Add any additional context, screenshots, or information that would help reviewers.


src/sap_cloud_sdk/core/auditlog_ng/__init__.py
```diff
@@ -85,16 +198,28 @@ def create_client(
...
+            resolution fails.
     """
     try:
         if config is None:
             try:
-                if not endpoint or not deployment_id or not namespace:
-                    raise ValueError(
-                        "endpoint, deployment_id, and namespace are required "
-                        "when config is not provided"
+                if destination_name and destination_instance:
```
betinacosta: I have my doubts if `destination_instace` should be mandatory. It does not seem necessary, but perhaps I am missing something. Any thoughts?

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
```diff
@@ -18,6 +19,16 @@
     )
     client = create_client(config=config)
 
+Usage — resolve from a Destination::
+
+    from sap_cloud_sdk.core.auditlog_ng import create_client
+
+    client = create_client(
+        destination_name="my-audit-destination",
+        destination_instance="my-binding-instance",
```
cassiofariasmachado, line 30: `destination_instance` could be optional also and we can use `default` by default

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
```diff
@@ -42,10 +53,92 @@
...
+    # Lazy import — keeps destination an optional dependency; importing auditlog_ng
+    # in environments without the destination package continues to work.
+    from sap_cloud_sdk.destination import (
+        ConsumptionOptions,
+        ConsumptionLevel,
+        create_client as _dest_create_client,
+    )
+
+    dest_client = _dest_create_client(instance=destination_instance)
+    options = ConsumptionOptions(fragment_name=fragment_name) if fragment_name else None
```
cassiofariasmachado: We must also inform the `fragment_level=ConsumptionLevel.SUBACCOUNT` in the `ConsumptionOptions`.

Ignore this comment if the default level is subaccount for this parameter.

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
```diff
@@ -61,13 +154,33 @@ def create_client(
...
+       via ``config``; all other keyword arguments are ignored.
+
+    2. **Destination-based resolution** — pass ``destination_name`` and
+       ``destination_instance`` (both required); ``fragment_name`` is optional.
+       The Destination module resolves the named destination at subaccount level
+       and extracts ``endpoint``, ``deployment_id`` (with fallback to
+       ``deploymentRegion``), and ``namespace`` from its properties.
+       The remaining keyword arguments (``cert_file``, ``key_file``, ``ca_file``,
+       ``insecure``, ``service_name``, ``batch``, ``compression``, ``schema_url``)
+       are still forwarded to the resulting :class:`AuditLogNGConfig`.
```
cassiofariasmachado: I think it is not necessary to list all the forwarded parameters here:
```suggestion
       The remaining keyword arguments are still forwarded to the resulting :class:`AuditLogNGConfig`.
```

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
cassiofariasmachado: We can expose it as a parameter, but take a look at my other comment

src/sap_cloud_sdk/core/auditlog_ng/user-guide.md
```diff
@@ -36,7 +37,31 @@ The client depends on generated protobuf classes.
...
+
+The destination must expose these custom properties:
+
+| Property           | Required | Description |
+|--------------------|----------|-------------|
+| `deploymentId`     | ✅ Yes (or `deploymentRegion`) | Deployment identifier. Falls back to `deploymentRegion` when absent or empty. |
+| `deploymentRegion` | ✅ Fallback | Used as `deployment_id` when `deploymentId` is missing or empty. |
+| `namespace`        | ✅ Yes   | Audit log namespace (e.g. `sap.als`). |
+
+The destination `url` is used as the OTLP gRPC endpoint. The lookup is always performed at subaccount level.
```
cassiofariasmachado: It is not always gRPC because it considers the environment variable `OTEL_EXPORTER_OTLP_PROTOCOL` to choose which protocol is used:
```suggestion
The destination `url` is used as the OTLP endpoint. The lookup is always performed at subaccount level.
```

You can take a look [here](https://github.com/SAP/cloud-sdk-python/blob/main/src/sap_cloud_sdk/core/auditlog_ng/client.py#L42).

Another approach we could follow is to use the `otlpProtocol` property (with the environment variable as a fallback) from the destination instead of using the environment variable when a `destination_name` is provided.

cassiofariasmachado (changes requested):
Nice work so far! I've added some comments

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
```diff
@@ -6,7 +6,8 @@
 The create_client() function accepts an AuditLogNGConfig and returns a
 ready-to-use AuditClient.
 
-Usage:
+Usage — explicit config::
```
NicoleMGomes: could we just add it as `Usage:` and then add comments, I believe this is how we usually do, but I can be worng

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
NicoleMGomes: I believe on agw we don't have destination name and destination_instance configuration, since this is padronized on runtime.

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
NicoleMGomes: Fragment Level is not on instance level? It should be specific to tenant if I remember correctly, and then we would need to ask this information too

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
```diff
@@ -42,10 +53,92 @@
...
+        name=destination_name, options=options, level=ConsumptionLevel.SUBACCOUNT
+    )
+
+    if destination is None:
+        raise ValueError(f"Destination '{destination_name}' was not found")
+
+    endpoint = destination.url
+    props = destination.properties
+
+    deployment_id = props.get(_DESTINATION_PROP_DEPLOYMENT_ID) or ""
```
NicoleMGomes: Maybe better to have an enum for this dest properties?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "sap-cloud-sdk"
-version = "0.26.1"
+version = "0.26.2"
```
NicoleMGomes: I don't think this is a minor since it's implementing new functionality

```suggestion
version = "0.27.0"
```

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
betinacosta: I tested passing the fragment with subaccount level and I got the necessary information back without errors. But I don't know if this will be the case for always. I will change to have subaccount as default.

cassiofariasmachado: It's tenant specific, but it was created in the subaccount level not in the instance level during the SPII. Either the destination and the fragment are being created in the subaccount level but tenant specific.

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
cassiofariasmachado: Do you think we should remove these parameters and consider them by default?

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
betinacosta: Fixed! Thanks.

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
betinacosta: Right now, the only "mandatory" parameter is `destination_name` if `endpoint`, `namespace` and `development_id` are not passed. Do you guys think it works this way?

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
betinacosta: I added the level to the fragment options, do you think any other information is required?

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
betinacosta: Fixed!

src/sap_cloud_sdk/core/auditlog_ng/user-guide.md
betinacosta: Fixed!

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
betinacosta: Fixed!

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
betinacosta: I think is better as well. Fixed.

pyproject.toml
betinacosta: It makes sense. Updated

src/sap_cloud_sdk/core/auditlog_ng/__init__.py
NicoleMGomes: I would keep equal to agw for consistency

NicoleMGomes (approved):

cassiofariasmachado (approved):

PR #192, opened 2026-07-01
fix(adms): align API surface with live ADMS wire format

ShiwaliKusum:

## Description

Aligns the ADMS Python SDK API surface with the live ADM service wire format. Several entity paths, key structures, and method signatures were incorrect — they compiled but failed at runtime (HTTP 400/404) against the real service.

**Configuration API fixes:**
- `DocumentTypeBusinessObjectTypeMap` uses a composite key (`DocumentTypeID` + `BusinessObjectNodeTypeUniqueID`), not a single UUID. `get_type_mapping`, `delete_type_mapping`, and `mark_default` updated accordingly.
- `FileExtensionPolicy` entity set renamed from `FileExtensionPolicy` to `DocumentTypeFileExtensionPolicy` (matches `$metadata`). Model fields corrected to `DocumentTypeID` + `FileExtension` composite key; `get_file_extension_policy(uuid)` removed as no such UUID key exists in ADM.
- `CreateDocumentTypeBoTypeMapInput` wire field names corrected from `DocumentTypeId`/`BusinessObjectNodeTypeUniqueId` to `DocumentTypeID`/`BusinessObjectNodeTypeUniqueID`.
- `MimeTypePolicy` removed from public `__init__` exports (no longer referenced by any public model).

**Document API fixes:**
- `update()` (sync + async): removed the implicit follow-up GET — the `UpdateDocument` action returns the partial response directly from ADM; callers who need the full document should call `get()` explicitly.
- `get_download_url()` (sync + async): `doc_content_version_id` is now optional (`None` → `DownloadDocument()` for latest version), matching the OData `Core.OptionalParameter` annotation.

**Job API fixes:**
- `StartJob` and `JobStatus` called without namespace prefix (unbound action/function imports do not use namespace prefix per OData spec).

## Related Issue

Closes #https://jira.tools.sap/browse/RBSADMS-47

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Configure credentials: `set -a && source .env.adms && set +a`
2. Run unit tests: `uv run pytest tests/adms/unit/ -v`
3. Run integration tests against a live ADM instance: `uv run pytest tests/adms/integration/ -v`
4. Verify `cfl`, `cmg`, `cmd`, `cgf`, `jz`, `js` commands work against the live tenant using `scripts/adms_cli.py`

## Checklist

- [x] All tests pass locally

## Breaking Changes

`get_type_mapping` and `delete_type_mapping` now require two arguments (`document_type_id`, `business_object_node_type_unique_id`) instead of one. However, the previous single-argument version accepted a `DocumentTypeBOTypeMapID` that the ADM service never supported — every call returned HTTP 400. Since the old signature was non-functional, this is treated as a bug fix rather than a breaking API change.

`get_file_extension_policy(uuid)` is removed. The ADM entity has no UUID primary key; policies are identified by `(DocumentTypeID, FileExtension)`. The old method would have always returned HTTP 404.

`update()` no longer performs an implicit GET after updating a document. The return value is now the partial action response from ADM (only changed fields). Callers requiring the full document after an update should call `documents.get()` explicitly.

## Additional Notes

All fixes were validated against a live ADM playground tenant. The `scripts/adms_cli.py` interactive CLI (kept locally, not included in this PR) was used for end-to-end verification of each endpoint.


scripts/adms_cli.py
```diff
@@ -0,0 +1,1611 @@
+#!/usr/bin/env python
+"""Interactive CLI for testing the ADMS SDK.
```
NicoleMGomes: Don't add scripts to sdk

src/sap_cloud_sdk/adms/_configuration_api.py
```diff
@@ -243,42 +244,42 @@ def create_type_mapping(
 
     @record_metrics(Module.ADMS, Operation.ADMS_CONFIG_GET_DOCTYPE_BOTYPE_MAP)
     def get_type_mapping(
-        self, document_type_bo_type_map_id: str
+        self, document_type_id: str, business_object_node_type_unique_id: str
```
NicoleMGomes: You are changing contract, this is a breaking change. Need to be careful when defining interface for consumers.

NicoleMGomes:
PR description is not following our template. Only raise PRs to us once it's complete.

src/sap_cloud_sdk/adms/_configuration_api.py
betinacosta, line 1: Several lines in this file seems to exceed 88 characters substantially, which is probably why pre-commit is failing.

src/sap_cloud_sdk/adms/_configuration_api.py
```diff
@@ -288,78 +325,106 @@ def get_all_file_extension_policies(
...
+            extra_headers=_subaccount_header(subaccount_id),
         )
 
 
+def _subaccount_header(subaccount_id: str | None) -> dict[str, str] | None:
+    """Build the X-SubaccountId header dict, or None if not provided."""
+    return {"X-SubaccountId": subaccount_id} if subaccount_id else None
+
+
 def _quote_guid(value: str) -> str:
```
betinacosta, line 428: It seems that this method is no longer used.

betinacosta:
The new `ApplicationTenant` operations introduce a `subaccount_id` parameter that routes to a per-tenant context, which is precisely the multi-tenancy routing pattern. The `user-guide.md` claims multi-tenancy is "Not Supported", but this PR adds `subaccount-scoped` operations — this is at minimum inconsistent and should be clarified. Additionally, the async `delete_application_tenant` (_configuration_api.py:766-769) does not pass `extra_headers=_subaccount_header(subaccount_id)` — the subaccount_id parameter is accepted but silently dropped on the async path.

betinacosta:
New methods `create_file_extension_policy`, `delete_file_extension_policy`, `get/create/delete ApplicationTenant`, and `mark_default` lack unit tests.

PR #221, opened 2026-07-09
feat(agent-memory): add create_checkpointer() factory with TTL support for LangGraph agents

santosh-nallur:
## Description

Adds a `create_checkpointer()` factory in the `agent_memory` module that returns a LangGraph `BaseCheckpointSaver` appropriate for the current environment. Enables LoB teams (e.g. IBD) to integrate LangGraph agent short-term memory without coupling their agent code to a specific checkpointer implementation.

**What ships in this PR:**

- `src/sap_cloud_sdk/agent_memory/factory/langgraph_checkpoint.py` — `create_checkpointer()` factory returning `InMemorySaver` (no TTL) or `TimedInMemorySaver` (with TTL)
- `src/sap_cloud_sdk/agent_memory/factory/_timed_memory.py` — `TimedInMemorySaver` extending `InMemorySaver` with a daemon background sweep thread for TTL-based thread eviction
- `src/sap_cloud_sdk/agent_memory/factory/__init__.py` — factory subpackage
- `tests/agent_memory/unit/test_langgraph_checkpointer.py` — factory unit tests
- `tests/agent_memory/unit/test_timed_memory.py` — `TimedInMemorySaver` unit tests
- `src/sap_cloud_sdk/agent_memory/user-guide.md` — LangGraph Checkpointer section

**usage:**

```python
from sap_cloud_sdk.agent_memory.factory.langgraph_checkpoint import create_checkpointer

# Without TTL
checkpointer = create_checkpointer()

# With TTL — evicts threads inactive for 1 hour
checkpointer = create_checkpointer(ttl_seconds=3600)

app = workflow.compile(checkpointer=checkpointer)
```

**`TimedInMemorySaver` design:**
- Tracks last-active time per thread via `_last_active` dict updated in `put()` / `aput()`
- Daemon sweep thread runs every 60 seconds and evicts threads exceeding `ttl_seconds` of inactivity
- Sweep is fully decoupled from the read/write path — hot path adds one dict write only
- `ttl_seconds` parameter stable for Phase 2 when `HanaAgentMemorySaver` ships

## Type of Change

- [x] Documentation update

## How to Test

```bash
# Run new unit tests
uv run pytest tests/agent_memory/unit/test_langgraph_checkpointer.py -v
uv run pytest tests/agent_memory/unit/test_timed_memory.py -v

# Run full unit suite
uv run pytest tests/agent_memory/unit/ -v

```

## Checklist

- [x] I have read the Contributing Guidelines
- [x] All tests pass locally
- [x] I have verified that my code follows the Code Guidelines
- [x] I have followed Conventional Commits for commit messages

## Breaking Changes

None. The `factory` subpackage is additive. No existing public API is changed.

## Additional Notes

- `TimedInMemorySaver` is a temporary home in the SDK factory subpackage. It will migrate to `langgraph-checkpoint-sap-agent-memory` (the separate checkpoint library in Phase 2 — no code change required for factory consumers.
- The `ttl_seconds` parameter is accepted on `create_checkpointer()` regardless of which backend is active, ensuring interface stability for Phase 2.
- Phase 2 factory (when `HanaAgentMemorySaver` ships) will detect the HANA binding at runtime and return it automatically


src/sap_cloud_sdk/agent_memory/user-guide.md
```diff
@@ -806,3 +816,84 @@ The `uaa` key must contain a JSON string with the XSUAA credentials:
...
+> Agent Memory Service is not yet supported.
+
+### Prerequisites
+
+`langgraph` must be installed. The SDK does not declare it as a dependency —
+you control your own LangGraph version.
+
+```bash
+pip install langgraph
+```
```
cassiofariasmachado: `langchain` is an optional dependency that we are using for the Agent Gateway module. We could follow the same approach here, adding `langgraph` as an optional dependency and validating and handling it with a proper message if the developer try to use the checkpointer without installing the `langgraph` dependency. Similar to what we did here:

- https://github.com/SAP/cloud-sdk-python/blob/main/src/sap_cloud_sdk/agentgateway/converters.py#L86
- https://github.com/SAP/cloud-sdk-python/blob/main/src/sap_cloud_sdk/agentgateway/converters.py#L16

src/sap_cloud_sdk/agent_memory/factory/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+
+Each module in this package provides a factory function that creates the
+appropriate framework-specific implementation for the current environment.
+
+Available factories:
+
+- :func:`sap_cloud_sdk.agent_memory.factory.langgraph_checkpoint.create_checkpointer`
+  — returns a LangGraph ``BaseCheckpointSaver`` (short-term memory)
+
+Usage::
```
NicoleMGomes:
```suggestion
Usage:
```

tests/agent_memory/unit/test_langgraph_checkpointer.py
```diff
@@ -0,0 +1,117 @@
+"""Unit tests for the create_checkpointer() LangGraph factory."""
```
NicoleMGomes, line 1: Better to keep same structure as src

santosh-nallur: Looking at the existing test layout across all modules (`tests/agentgateway/unit/`, `tests/destination/unit/`, etc.), the convention is a flat list of test files under unit/ regardless of whether the source has subpackages. For example `agentgateway/converters.py` lives in a subpackage but its test is `tests/agentgateway/unit/test_converters.py` — not `tests/agentgateway/unit/converters/test_converters.py`.

Following that pattern, `test_langgraph_checkpointer.py` and `test_timed_memory.py` are in the right place under `tests/agent_memory/unit/`.

Could you clarify if you meant something else, like a different naming or structure convention?

src/sap_cloud_sdk/agent_memory/user-guide.md
santosh-nallur: Thanks. I had it before, but removed it. I had an in-correct impression `pip install sap-cloud-sdk[langgraph]` would always restrict them to langgraph version we specify under optional section. But looking at it again, seems not the case because of bounds we use and additionally the agent devs could still specify a specific  version.

santosh-nallur:
Thanks @NicoleMGomes , @cassiofariasmachado, I realised that the new TTL field for checkpointer could be a configuration via decorators. Rather than introducing a new checkpointer specific decorator, reused the existing agent_config instead. Added an example in the [user-guide.md ](https://github.com/santosh-nallur/cloud-sdk-python/blob/feat/local-factory/src/sap_cloud_sdk/agent_memory/user-guide.md#exposing-ttl-as-a-configurable-parameter-with-agent_config)
Hope this is fine ?

tests/agent_memory/unit/test_langgraph_checkpointer.py
NicoleMGomes: You are right, I will review it and if needed, we will do adjustments later. Thanks

pyproject.toml
```diff
@@ -33,6 +34,7 @@ dependencies = [
 extensibility = ["a2a-sdk>=0.2.0"]
 starlette = ["starlette>=0.40.0"]
 langchain = ["langchain-core>=1.2.7"]
+langgraph = ["langgraph>=1.0.0"]
```
NicoleMGomes, line 36: @cassiofariasmachado do we need to make anything to install it conditionally to langchain install?

cassiofariasmachado: It's already handled as an optional dependency like `langchain`.

cassiofariasmachado (dismissed):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "sap-cloud-sdk"
-version = "0.33.2"
+version = "0.32.0"
```
cassiofariasmachado: Please bump the version to `0.34.0`
```suggestion
version = "0.34.0"
```

cassiofariasmachado (approved):

NicoleMGomes (approved):

PR #224, opened 2026-07-14
feat(telemetry): warn at startup when required env vars are missing

jeanscherf:
## Summary

- `_get_system_role()`: emits `logger.warning()` when `APPFND_CONHOS_SYSTEM_ROLE` is unset, instead of silently defaulting to `unknown`
- `_get_ord_id()`: emits `logger.warning()` when `ORD_DOCUMENT_ID` is unset, instead of silently omitting `sap.ord.id` from the OTel resource
- Follows the existing `logging.getLogger(__name__)` convention used throughout the telemetry module

## Test plan

- [ ] Run with env vars unset — confirm both warnings appear in application logs at startup
- [ ] Run with env vars set — confirm no warnings are emitted
- [ ] Existing unit tests in `tests/core/unit/telemetry/test_config.py` pass unchanged

### DIFF
--- a/src/sap_cloud_sdk/core/telemetry/config.py
+++ b/src/sap_cloud_sdk/core/telemetry/config.py
@@ -1,5 +1,6 @@
 """Configuration for OpenTelemetry telemetry."""
 
+import logging
 import os
 from dataclasses import dataclass
 from typing import Optional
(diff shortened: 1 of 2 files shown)


LucasAlvesSoares (approved):

src/sap_cloud_sdk/core/telemetry/config.py
```diff
@@ -77,7 +80,16 @@ def _get_system_role() -> str:
         System role from APPFND_CONHOS_SYSTEM_ROLE environment variable,
         or "ZAFT" if not set.
     """
-    return os.getenv(ENV_SYSTEM_ROLE, DEFAULT_UNKNOWN)
+    value = os.getenv(ENV_SYSTEM_ROLE)
+    if not value:
+        logger.warning(
+            "APPFND_CONHOS_SYSTEM_ROLE is not set. "
```
gpretto, line 86: APPFND_CONHOS_SYSTEM_ROLE still the name? I thought, we suposed to remove the "conhos" from naming.

Should we add something like: "review app.yaml env object" or something like it?

jeanscherf:
Closing — system role will be injected automatically from the blueprint onboarding file in CI. SDK-level warnings are not needed.

PR #231, opened 2026-07-15
feat:add new secret resolver to modules

betinacosta:
## Description

This PR implements the secret resolver into the sdk modules.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Code refactoring
- [ ] Dependency update

## How to Test
```python
from sap_cloud_sdk import (
    adms,
    agent_memory,
    agentgateway,
    destination,
    dms,
    objectstore,
    print as print_service
)

from sap_cloud_sdk.core import (
    auditlog_ng,
    data_anonymization
)


def main() -> None:
    # Destination client
    destination_client = destination.create_client()
    print(f"Destination: {destination_client}")

    # ADMS client
    adms_client = adms.create_client()
    print(f"ADMS: {adms_client}")

    # Agent Gateway
    agent_gateway_client = agentgateway.create_client()
    print(f"Agent Gateway: {agent_gateway_client}")

    # Agent Memory
    agent_memory_client = agent_memory.create_client()
    print(f"Agent Memory: {agent_memory_client}")

    # DMS
    dms_client = dms.create_client()
    print(f"DMS: {dms_client}")

    # Object Store
    object_store_client = objectstore.create_client("default")
    print(f"Object Store: {object_store_client}")

    # Print Service
    print_client = print_service.create_client()
    print(f"Print: {print_client}")

    # Auditlog V3
    auditlog_client = auditlog_ng.create_client()
    print(f"Auditlog NG: {auditlog_client}")

    # Data Anonymization
    data_client = data_anonymization.create_client()
    print(f"Data Anonymization: {data_client}")

    

if __name__ == "__main__":
    main()
```

## Checklist

- [ ] I have verified that my changes solve the issue
- [ ] All tests pass locally
- [ ] I have verified that my code follows the [Code Guidelines](../docs/GUIDELINES.md)
- [ ] My code does not contain sensitive information (credentials, tokens, etc.)
- [ ] I have followed [Conventional Commits](https://www.conventionalcommits.org/) for commit messages

## Breaking Changes

If this PR introduces breaking changes, please describe:

- What breaks
- Migration path for users
- Alternative approaches considered

## Additional Notes

Add any additional context, screenshots, or information that would help reviewers.


src/sap_cloud_sdk/core/secret_resolver/__init__.py
NicoleMGomes: This method shouldn't exist anymore

src/sap_cloud_sdk/core/secret_resolver/_resolvers.py
NicoleMGomes: Maybe instead of module, could it be `service_name` and `instance_name`?

src/sap_cloud_sdk/core/secret_resolver/env_resolver.py
NicoleMGomes: Could be moved to resolve.

I'm just thinking this is not an EnvVarResolver generic, because it expects module and instance. Should we maybe have an SdkEnvVarResolver and an EnvVarResolver just with the variable name?

src/sap_cloud_sdk/core/secret_resolver/mount_resolver.py
NicoleMGomes: Same points here.

src/sap_cloud_sdk/core/secret_resolver/sdk_config.py
NicoleMGomes, line 51: I believe this configuration could be global or per module.

NicoleMGomes: Aren't we missing an `BindingRootResolver` indeed?

NicoleMGomes:
```suggestion
def get() -> Optional[SdkConfig]:
```
No need to duplicate sdk_config in the name since you are already have a class.

src/sap_cloud_sdk/core/secret_resolver/_resolvers.py
NicoleMGomes: Maybe we will need to have a generic list of args or an interface each resolver implement for ResolverParameters.

src/sap_cloud_sdk/core/data_anonymization/config.py
```diff
@@ -111,7 +112,7 @@ def extract_config(self) -> DataAnonymizationConfig:
         )
 
 
-def _load_config_from_env(instance: str = "default") -> DataAnonymizationConfig:
+def _load_secrets(instance: str = "default") -> DataAnonymizationConfig:
```
NicoleMGomes, line 115: I believe this encapsulation is not needed. We could also have an central enum for module names if we need it, instead of magic strings. What do you think?

src/sap_cloud_sdk/core/secret_resolver/_resolvers.py
betinacosta: I think `service_name` is better than `module`,  but we already use `instance` in other situations, so I believe is better to keep it for consistence.

src/sap_cloud_sdk/core/secret_resolver/env_resolver.py
betinacosta: Makes sense. I will move `_load_from_env` to resolve. I will also rename to `SdkEnvVarResolver`. Regarding having a generic `EnvVarResolver` just with the variable name, do we have an use case for it?

src/sap_cloud_sdk/core/secret_resolver/sdk_config.py
betinacosta: I renamed the `BindingResolver` to `Resolver`.

src/sap_cloud_sdk/core/data_anonymization/config.py
betinacosta: I like the idea of having an enum, but I would leave the encapsulation for better organization, so in the client we can do only: `binding = config or load_secrets(instance)`

src/sap_cloud_sdk/core/secret_resolver/_resolvers.py
betinacosta: I'm not sure I follow. We already have a Protocol class called `Resolver` that the other resolvers implement.

src/sap_cloud_sdk/core/data_anonymization/config.py
betinacosta: I accept suggestions on where to have this enum.

src/sap_cloud_sdk/core/secret_resolver/mount_resolver.py
cassiofariasmachado, line 1: I believe that we will have more and more resolvers, maybe we should move them to a folder named `resolvers`. What do you think?

PR #240, opened 2026-07-21
feat(telemetry): decouple module/operation identifiers from enums

jeanscherf:
## Description

The `record_metrics` decorator and the underlying `record_request_metric`, `record_error_metric`, and `default_attributes` functions previously required `Module` and `Operation` enum values defined inside this package. This made it impossible for external packages to use the OSS telemetry infrastructure without adding entries to those enums via a PR to this repo.

This change relaxes the type constraints from `Module` / `Operation` enums to plain `str` throughout. Since `Module` and `Operation` already extend `str`, all existing callsites are unaffected — enum values continue to work exactly as before. External packages can now pass plain strings (e.g. `@record_metrics("my_module", "my_operation")`) and plug into the same OTel infrastructure without any changes here.

The `isinstance(source_kwarg, Module)` guard in the decorator is also relaxed to `isinstance(source_kwarg, str)` for the same reason.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Use the decorator with plain strings from outside the package:
   ```python
   from sap_cloud_sdk.core.telemetry import record_metrics

   class MyClient:
       @record_metrics("my_module", "my_operation")
       def my_method(self):
           ...
   ```
2. Verify no type errors are raised and metrics are emitted normally when `OTEL_EXPORTER_OTLP_ENDPOINT` is set.
3. Verify existing enum usage (`Module.DESTINATION`, `Operation.GET_DESTINATION`) continues to work unchanged.
4. Run `uv run pytest tests/core/unit/telemetry/test_metrics_decorator.py` — all tests including the new `TestRecordMetricsDecoratorPlainStrings` class should pass.

## Checklist

- [x] All tests pass locally

## Additional Notes

The `Module` and `Operation` enums remain the canonical, recommended way to call the API for OSS consumers — they are still exported and fully supported. This change only opens the contract so external packages are not forced to contribute to this repo to use the telemetry layer.

### DIFF
--- a/src/sap_cloud_sdk/core/telemetry/metrics_decorator.py
+++ b/src/sap_cloud_sdk/core/telemetry/metrics_decorator.py
@@ -4,7 +4,6 @@ from functools import wraps
 from typing import Callable, Optional, TypeVar, ParamSpec
 
 from sap_cloud_sdk.core.telemetry import (
-    Module,
     record_request_metric,
     record_error_metric,
 )
@@ -14,7 +13,7 @@ R = TypeVar("R")
 
 
 def record_metrics(
-    module: Module, operation: str, deprecated: bool = False
+    module: str, operation: str, deprecated: bool = False
 ) -> Callable[[Callable[P, R]], Callable[P, R]]:
     """Decorator to automatically record request and error metrics for SDK operations.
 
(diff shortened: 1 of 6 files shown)


src/sap_cloud_sdk/core/telemetry/metrics_decorator.py
```diff
@@ -14,7 +13,7 @@
 
 
 def record_metrics(
-    module: Module, operation: str, deprecated: bool = False
+    module: str, operation: str, deprecated: bool = False
```
cassiofariasmachado, line 16: Couldn’t it be `str | Module`?

jeanscherf: Module is an extension of str. We could do it but it would be redundant.
I thought we should be honest. 😉 
We still have the docstring and the enum exports.

cassiofariasmachado (approved):

PR #242, opened 2026-07-22
feat(auditlog_ng): auto-inject tenant and user from IAS request context

tiagoek:
## Summary

- Adds `ias/_context.py` with a `ContextVar`-backed `set_auth_context` / `get_auth_context` to store IAS claims per async request
- Refactors `StarletteIASTelemetryMiddleware` to call `set_auth_context(claims)` on every request via a unified `_parse_request()` helper
- Wires `AuditClient.send()` to auto-fill `common.tenant_id` (from `app_tid`) and `common.user_initiator_id` (from `user_uuid`) before protovalidate runs — only when not already set by the caller
- Documents the new automatic injection behaviour in `auditlog_ng/user-guide.md`

## Test plan

- [ ] Verify `_fill_common_from_auth_context` fills empty fields from a mocked IAS context
- [ ] Verify explicit caller values are not overwritten
- [ ] Verify no injection when no auth context is set (unauthenticated request)
- [ ] Verify `send()` still raises `ValidationError` when `tenant_id` is absent and no context is set

### DIFF
--- a/src/sap_cloud_sdk/core/auditlog_ng/client.py
+++ b/src/sap_cloud_sdk/core/auditlog_ng/client.py
@@ -37,6 +37,21 @@ from sap_cloud_sdk.core.auditlog_ng.config import (
 from sap_cloud_sdk.core.auditlog_ng.exceptions import ValidationError
 from sap_cloud_sdk.core.telemetry import Module, Operation, record_metrics
 from sap_cloud_sdk.core.telemetry.config import ENV_OTLP_PROTOCOL
+from sap_cloud_sdk.ias._context import get_auth_context
+
+
+def _fill_common_from_auth_context(event: Message) -> None:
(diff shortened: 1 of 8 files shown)


src/sap_cloud_sdk/core/auditlog_ng/client.py
```diff
@@ -37,6 +37,21 @@
...
+
+
+def _fill_common_from_auth_context(event: Message) -> None:
+    """Back-fill tenant_id and user_initiator_id from the request auth context."""
+    claims = get_auth_context()
+    if claims is None:
+        return
+    common = getattr(event, "common", None)
+    if common is None:
+        return
```
LucasAlvesSoares: maybe if common is None we create it?

LucasAlvesSoares (approved):

PR #256, opened 2026-08-04
feat(aicore): reactive credential reload and proactive secret watcher

tiagoek:
## Description

Addresses credential rotation failures in long-running agent pods.

When BTP rotates a service binding, kubelet updates the mounted secret volume within ~2 minutes. However, agent pods are never restarted — LiteLLM's cached OAuth token (up to 12h lifetime) eventually expires and the refresh attempt fails because the process still holds the old `client_secret`. This PR makes credential rotation transparent to agents without requiring pod restarts.

### 1. Reactive credential reload on `AuthenticationError`

`completion()` and `acompletion()` intercept `litellm.AuthenticationError`, reload credentials from the mounted secret volume via `set_aicore_config()`, and retry the call once. If the retry also fails, the error propagates normally.

### 2. Proactive secret watcher (`watch_aicore_config`)

A daemon thread polls the secret directory mtime every 60 seconds. On change (kubelet performs an atomic symlink swap on rotation), calls `set_aicore_config()` proactively — before the cached OAuth token expires. Avoids the 401 entirely rather than recovering from it.

Typical usage:

```python
from sap_cloud_sdk.aicore import set_aicore_config, watch_aicore_config

set_aicore_config()
watch_aicore_config()   # starts daemon thread — call once at startup
```

### 3. LangGraph / ChatLiteLLM compatibility (`patch_litellm_for_credential_rotation`)

LangGraph agents use `ChatLiteLLM` which calls `litellm.completion` directly, bypassing our `completion()` wrapper. `patch_litellm_for_credential_rotation()` wraps `litellm.completion` and `litellm.acompletion` at the module level so all callers get transparent reload on 401 — including `ChatLiteLLM`.

```python
from sap_cloud_sdk.aicore import (
    set_aicore_config,
    patch_litellm_for_credential_rotation,
    watch_aicore_config,
)

set_aicore_config()
patch_litellm_for_credential_rotation()   # for LangGraph / ChatLiteLLM agents
watch_aicore_config()
```

## Related Issues

- Stacked on: this PR is the base; #271 stacks on top

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

```bash
python -m pytest tests/aicore/unit/ -v
# 111 passed
```

Key test files:
- `tests/aicore/unit/test_aicore_watcher.py` — proactive watcher (8 tests)
- `tests/aicore/unit/test_credential_rotation_flow.py` — env update contract (6 tests)
- `tests/aicore/unit/test_completion.py` — reactive reload (sync + async)
- `tests/aicore/unit/test_langgraph_compat.py` — LangGraph / ChatLiteLLM compatibility (15 tests)

## Checklist

- [x] All tests pass locally

## Breaking Changes

None. All changes are additive:

- `watch_aicore_config()` — new public function, opt-in
- `patch_litellm_for_credential_rotation()` — new public function, opt-in, idempotent
- Retry on `AuthenticationError` in `completion()` — same exception type propagates if retry also fails; no contract change


src/sap_cloud_sdk/aicore/completion.py
```diff
@@ -39,12 +49,31 @@
...
 from typing import Any
 
 import litellm
 
 from .filtering.filters import _parse_input_filter_error
 
+logger = logging.getLogger(__name__)
+
+
+def reload_aicore_credentials() -> None:
```
NicoleMGomes, line 62: Why are we creating a new method that only calls other?

src/sap_cloud_sdk/aicore/__init__.py
```diff
@@ -30,6 +30,16 @@
 
 logger = logging.getLogger(__name__)
 
+# When set, the infrastructure sidecar adds the mTLS certificate transparently.
+# The SDK calls the XSUAA token endpoint over plain HTTPS with only client_id.
+# No client_secret or certificate material is required in the service binding.
+TRANSPARENT_TLS_ENV_VAR = "AICORE_TRANSPARENT_TLS"
```
NicoleMGomes, line 36: Can we understand if we can have a single variable to set transparent proxy usage and not specific by module?

src/sap_cloud_sdk/aicore/completion.py
tiagoek: reload_aicore_credentials() serves two purposes: it's called automatically by completion()/acompletion() on AuthenticationError (reactive reload on credential rotation), and it's also exposed as a public API for callers that need to trigger a manual reload. Keeping it as a named function makes the automatic behavior explicit and gives callers a stable surface without coupling them to set_aicore_config() internals.

NicoleMGomes: I still disagree in having an new function just to wrapper with a new nomenclature. If in future you believe more will be needed, it's ok.

src/sap_cloud_sdk/aicore/__init__.py
NicoleMGomes: This probably could be related to secrets resolver refactor.

tiagoek:
## Production validation — proactive credential reload (`watch_aicore_config`)

### Test suite (cloud-sdk-python)

All 70 unit tests pass on branch `feat/aicore-transparent-tls`:

```
tests/aicore/unit/test_aicore_watcher.py          8 passed
tests/aicore/unit/test_credential_rotation_flow.py 6 passed
tests/aicore/unit/test_aicore.py                  (existing, green)
tests/aicore/unit/test_completion.py              (existing, green)
```

### Integration into autonomous-documentation-org

The vendor shim (`vendor/sap_cloud_sdk_aicore/`) was synced with `watch_aicore_config` + `_get_secret_dir_mtime`.

**Finding (critical):** Starlette 1.3.1 does NOT propagate lifespan events to mounted sub-apps — verified with a live ASGI simulation:

```python
# Starlette 1.3.1 — Mount does not trigger FastAPI sub-app lifespan
outer = Starlette(routes=[Mount('/sub', app=fastapi_sub_app)])
# result: started = []  (sub-app _lifespan never called)
```

`watch_aicore_config()` was therefore moved to `app/main.py` at module level alongside `set_aicore_config()` — the same proven path that loads credentials on every boot.

### Production proof (Kyma managed runtime, `auto-doc-dev-eu12`)

Thread count in PID 1 before and after deploying commit `a72f5e0`:

| Pod | Image | Threads in PID 1 |
|-----|-------|-----------------|
| Pre-fix (`6ad5b0854983`) | `0.3.3-20260825...` (no watcher call) | **14** |
| Post-fix (`1858f9060d95`) | `0.3.3-20260825200836_fd446f2+1` | **15** |

The extra thread is `aicore-secret-watcher` (daemon, polls `/etc/secrets/appfnd/aicore/aicore-instance/` mtime every 30 s). It will call `set_aicore_config()` on the next kubelet symlink-swap rotation without requiring a pod restart.

```
autonomous-documentation-org unit tests: 286 passed, 1 skipped — coverage 90.51%
```

tiagoek:
## E2E validation — reactive credential reload (live pod test)

**Environment:** Kyma managed runtime, namespace `auto-doc-dev-eu12`, pod `autonomous-documentation-ee13df69-1858f9060d95` (image `0.3.3-20260825`)

### Test script

[`tests/integration/e2e_llm_reactive_reload.py`](https://github.tools.sap/app-fnd-autonomous-documentation/autonomous-documentation/blob/main/tests/integration/e2e_llm_reactive_reload.py) — run via `kubectl exec` in the pod where the secret volume is mounted.

### Execution output

```
[1] Real secret loaded from file: 263a99c5...
[2] Env poisoned with invalid secret
INFO sap_cloud_sdk_aicore.completion: AI Core credentials reloading after authentication failure
INFO sap_cloud_sdk_aicore: Loaded AICORE_CLIENT_SECRET from file: /etc/secrets/appfnd/aicore/aicore-instance/clientsecret
INFO sap_cloud_sdk_aicore: AI Core configuration has been set successfully
[3] Post-call secret: (cleared by _clear_client_secret — expected PR#257)
[4] LLM response: 'OK'
[OK] Test 1 PASSED — reactive reload intercepted the failure, retried, LLM responded
[5] LiteLLM aicore/client namespace attrs: [...in_memory_llm_clients_cache...]
[6] AICORE_CLIENT_SECRET in env: (absent — cleared after token acquisition)
[OK] Test 2 PASSED — os.environ is the sole credential store; no per-model client object
INFO sap_cloud_sdk_aicore.completion: AICORE_CLIENT_SECRET cleared from environment after token acquisition (AFSDK-4291)
```

### What this proves

| Scenario | Result |
|----------|--------|
| Invalid `AICORE_CLIENT_SECRET` in env → LiteLLM call | `AuthenticationError` caught by handler |
| `reload_aicore_credentials()` reads from mounted file | All 5 AICORE_* vars reloaded from `/etc/secrets/appfnd/aicore/aicore-instance/` |
| Retry with restored credentials | LLM responded `'OK'` |
| PR#257 `_clear_client_secret()` behaviour | Secret cleared from env after token acquisition — confirmed expected |
| LiteLLM `in_memory_llm_clients_cache` | Memory cache keyed by model identifier, not by credential values — `os.environ` read fresh on every OAuth token request |

src/sap_cloud_sdk/aicore/completion.py
tiagoek: Removed!

src/sap_cloud_sdk/aicore/__init__.py
tiagoek: Improved! However in future needs to be refactoring using new secret resolver.

src/sap_cloud_sdk/aicore/__init__.py
```diff
@@ -175,8 +176,73 @@ def set_aicore_config(instance_name: str = "aicore-instance") -> None:
...
+    secret_dir = os.path.join(resolve_base_mount(), "aicore", instance_name)
+    try:
+        return os.stat(secret_dir).st_mtime
+    except OSError:
+        return 0.0
+
+
+def watch_aicore_config(
+    instance_name: str = "aicore-instance",
+    interval: float = 30.0,
```
jeanscherf: If the token lifetime is 12h, could we maybe have a less conservative interval, e.g. say 60s, 300s or something? Though I think 30s is totally harmless as well. Your call.

jeanscherf: Also, is this a float because of the fast test execution?

tiagoek: good point! increasing to avoid problems!

tiagoek: yes, considering is possible to test with 0.5. this is intentional.

pedro-hca (approved):
Image

tests/aicore/unit/test_completion.py
thiagob, line 1: Would be possible to add a unit test with lang graph? 

Checking the [template](https://github.tools.sap/btp-ai/ibd/blob/main/skills/agent/templates/app/agent_executor.py) I don't see our completion or acompletion being used. Would be important to map if there is any in the agent itself that LoBs will have to do, or if it is just a simper version bump.

tiagoek: improving!

tiagoek: LangGraph agents use ChatLiteLLM which calls litellm.completion directly and bypasses our completion() wrapper, so the reactive 401 handler wouldn't fire for them

To cover that gap I added patch_litellm_for_credential_rotation(), it wraps litellm.completion and litellm.acompletion at the module level so ALL callers (including ChatLiteLLM) get the same transparent reload on 401. Combined with watch_aicore_config(), LangGraph agents are fully covered

src/sap_cloud_sdk/aicore/__init__.py
```diff
@@ -175,8 +176,132 @@ def set_aicore_config(instance_name: str = "aicore-instance") -> None:
...
+def _get_secret_dir_mtime(instance_name: str = "aicore-instance") -> float:
+    """Return the mtime of the AI Core secret directory, or 0.0 if it does not exist."""
+    secret_dir = os.path.join(resolve_base_mount(), "aicore", instance_name)
+    try:
+        return os.stat(secret_dir).st_mtime
+    except OSError:
+        return 0.0
+
+
+def watch_aicore_config(
```
betinacosta, line 198: Question: Shouldn't this have a `@record_metrics`?

src/sap_cloud_sdk/aicore/__init__.py
```diff
@@ -175,8 +176,132 @@ def set_aicore_config(instance_name: str = "aicore-instance") -> None:
...
+                    last_mtime = current_mtime
+            except Exception:
+                logger.exception("Error during proactive AI Core credential reload")
+
+    thread = threading.Thread(target=_watch, daemon=True, name="aicore-secret-watcher")
+    thread.start()
+    return thread
+
+
+def patch_litellm_for_credential_rotation() -> None:
```
betinacosta, line 254: Same question here

src/sap_cloud_sdk/aicore/__init__.py
tiagoek: good catch! improving now

src/sap_cloud_sdk/aicore/__init__.py
tiagoek: good catch! improving now

betinacosta (dismissed):

betinacosta (dismissed):

betinacosta (dismissed):

src/sap_cloud_sdk/aicore/user-guide.md
```diff
@@ -47,6 +47,81 @@ set_aicore_config(instance_name="aicore-production")
...
+```python
+import threading
+from sap_cloud_sdk.aicore import set_aicore_config, watch_aicore_config
+
+set_aicore_config()          # load credentials at startup
+watch_aicore_config()        # proactive reload on secret rotation
+
+# Optional: stop cleanly at shutdown
+_stop = threading.Event()
+watch_aicore_config(stop_event=_stop)
```
jeanscherf: maybe we should add super clear that this is an alternative? 
Or I feel people will call the watcher twice. ;)

src/sap_cloud_sdk/aicore/__init__.py
```diff
@@ -175,8 +176,134 @@ def set_aicore_config(instance_name: str = "aicore-instance") -> None:
...
+                if current_mtime != last_mtime:
+                    logger.info(
+                        "AI Core secret volume changed — proactively reloading credentials"
+                    )
+                    set_aicore_config(instance_name=instance_name)
+                    last_mtime = current_mtime
+            except Exception:
+                logger.exception("Error during proactive AI Core credential reload")
+
+    thread = threading.Thread(target=_watch, daemon=True, name="aicore-secret-watcher")
```
jeanscherf, line 248: does it make sense to extract that aicore-secret-watcher to a constant? I saw it only once so maybe the string is fine. Your call.

src/sap_cloud_sdk/aicore/__init__.py
```diff
@@ -175,8 +176,134 @@ def set_aicore_config(instance_name: str = "aicore-instance") -> None:
...
+def _get_secret_dir_mtime(instance_name: str = "aicore-instance") -> float:
+    """Return the mtime of the AI Core secret directory, or 0.0 if it does not exist."""
+    secret_dir = os.path.join(resolve_base_mount(), "aicore", instance_name)
+    try:
+        return os.stat(secret_dir).st_mtime
+    except OSError:
+        return 0.0
+
+
+@record_metrics(Module.AICORE, Operation.AICORE_WATCH_CONFIG)
```
jeanscherf: Are we measuring this only once (watcher is started!)? Because this will always be 1/pod.
the `AICORE_SET_CONFIG` for me is way more useful. Not sure if this here brings any value to be honest.

src/sap_cloud_sdk/aicore/__init__.py
```diff
@@ -175,8 +176,134 @@ def set_aicore_config(instance_name: str = "aicore-instance") -> None:
...
+    def _watch() -> None:
+        nonlocal last_mtime
+        while not stop_event.wait(timeout=interval):
+            try:
+                current_mtime = _get_secret_dir_mtime(instance_name)
+                if current_mtime != last_mtime:
+                    logger.info(
+                        "AI Core secret volume changed — proactively reloading credentials"
+                    )
+                    set_aicore_config(instance_name=instance_name)
```
jeanscherf: Maybe we could get richer metrics if we instrument manually:
something like

```
# operation.py
AICORE_SET_CONFIG         = "set_aicore_config"
AICORE_PROACTIVE_RELOAD   = "aicore_proactive_reload"   # new
AICORE_REACTIVE_RELOAD    = "aicore_reactive_reload"    # new

# __init__.py -> watcher calls a labelled wrapper instead of set_aicore_config directly
def _reload_proactive(instance_name: str) -> None:
    set_aicore_config(instance_name=instance_name)

_reload_proactive = record_metrics(Module.AICORE, Operation.AICORE_PROACTIVE_RELOAD)(_reload_proactive)

# completion.py -> reactive path
@record_metrics(Module.AICORE, Operation.AICORE_REACTIVE_RELOAD)
def _reload_reactive() -> None:
    set_aicore_config()
```

jeanscherf: Let me know what you think about it.
We could even delete the decorator in the set_aicore_config() then so we don't get double metrics.
But that is totally optional

src/sap_cloud_sdk/aicore/user-guide.md
tiagoek: I split into two separate blocks with a clear label "alternative form — pick one" so it's obvious they're mutually exclusive

src/sap_cloud_sdk/aicore/__init__.py
tiagoek: Used only once, leaving it inline. Agreed it's fine as-is

src/sap_cloud_sdk/aicore/__init__.py
tiagoek: that metric had little value. 
Removed it and replaced with the approach you suggested in comment 4, see below

src/sap_cloud_sdk/aicore/__init__.py
tiagoek: Great suggestion, done! Added AICORE_PROACTIVE_RELOAD and AICORE_REACTIVE_RELOAD to the Operation enum. The watcher now calls _reload_proactive() (decorated with the new op) and both reactive paths, completion.py and patch_litellm, call _reload_reactive(). So every actual rotation event shows up in telemetry, both proactive and 401-triggered

tests/aicore/unit/test_completion.py
tiagoek: Yes, we tested it — the autonomous documentation agent ran an E2E A2A test today (2026-08-31) and it confirmed your concern.

Gap found: A2A uses ChatLiteLLM (LangGraph), which calls litellm.completion directly — bypassing the SDK's completion() wrapper. So the reactive handler in PR #256 (which lives in the wrapper) does not fire for that path.

Two fixes available in PR #256:

watch_aicore_config() — proactive watcher, covers A2A with zero agent code changes (reloads credentials before the OAuth token expires, so no 401 ever reaches the agent)
patch_litellm_for_credential_rotation() — also in PR #256, patches litellm.completion globally so even the ChatLiteLLM path gets transparent reactive reload
The E2E test (e2e_a2a_reactive_reload.py) is now in the autonomous-documentation-org integration tests, runnable inside the Kyma pod. The recommended fix for A2A agents is just watch_aicore_config() at startup — no agent code changes required.

jeanscherf (approved):

betinacosta (approved):

PR #260, opened 2026-08-07
feat: extensibility ums destination changes

hariturangi:
## Description
UMS Destination - New connectivity model 

This change is needed because the AppFND team suggested using an existing destination name (sap-managed-runtime-ias-${landscape}) instead of creating a new destination name (sap-managed-runtime-ums-${landscape}) to read the extensions.
Hence, the current code needs to be modified to use the one the AppFND team suggested.

Environment variable support: 
Added APPFND_CONHOS_UMS_URL to enable the new IAS-based flow

Dynamic destination resolution:
When APPFND_CONHOS_UMS_URL is set → uses IAS destination (sap-managed-runtime-ias-{landscape})
Otherwise → falls back to legacy UMS destination (sap-managed-runtime-ums-{landscape})

URL resolution logic:
New flow: Base URL from APPFND_CONHOS_UMS_URL environment variable

Legacy flow: Base URL from destination configuration
Both flows: mTLS certificate always from the resolved destination

## Related Issue

NA
(Link to the GitHub issue this PR addresses)

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [X] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

Describe how reviewers can test your changes:
1. Step 1
2. Step 2
3. Expected result

## Checklist

- [X] I have added/updated automated tests to cover my changes
- [X] All tests pass locally
- [X] I have verified that my code follows the [Code Guidelines](../docs/GUIDELINES.md)
- [X] I have updated documentation (if applicable)
- [X] I have added type hints for all public APIs
- [X] My code does not contain sensitive information (credentials, tokens, etc.)
- [X] I have followed [Conventional Commits](https://www.conventionalcommits.org/) for commit messages

## Breaking Changes
There are no breaking changes. We have BCT (Backwards Compatibility) enabled; if the env variable is not found, then it will go with old behaviour

## Additional Notes

Tested in the local agent and it worked.


src/sap_cloud_sdk/extensibility/_ums_transport.py
```diff
@@ -53,7 +53,9 @@
 
 ENV_CONHOS_LANDSCAPE = "APPFND_CONHOS_LANDSCAPE"
 ENV_UMS_DESTINATION_NAME = "APPFND_UMS_DESTINATION_NAME"
+ENV_UMS_URL = "APPFND_CONHOS_UMS_URL"
 _UMS_DESTINATION_PREFIX = "sap-managed-runtime-ums-"
```
NicoleMGomes: I believe we can remove the ums destination since this was never provisioned on managed runtime

NicoleMGomes: Better to not keep the two ways so code is more readable

NicoleMGomes:
How have you validated these changes? On which agent?

hariturangi:
> How have you validated these changes? On which agent?

We have an agent https://github.tools.sap/agent-extensibility-test/extensibility-tools-agent; I have started this agent locally and applied the SDK changes to this agent

src/sap_cloud_sdk/extensibility/_ums_transport.py
hariturangi: There are some existing agents that are referring to this destination, which are manually created

NicoleMGomes: Ok, but the new approach should work for all agents since it's automated, no need to keep this work-around on sdk. And if the new approach is not there, they don't need to adopt the new version right now.

hariturangi: Modified the code as suggested

src/sap_cloud_sdk/extensibility/_ums_transport.py
```diff
@@ -53,7 +53,8 @@
 
 ENV_CONHOS_LANDSCAPE = "APPFND_CONHOS_LANDSCAPE"
 ENV_UMS_DESTINATION_NAME = "APPFND_UMS_DESTINATION_NAME"
```
NicoleMGomes: Do we still need this variable? As far as I know this is not being set

src/sap_cloud_sdk/extensibility/_ums_transport.py
```diff
@@ -556,13 +578,28 @@ def get_extension_capability_implementation(
...
+        #    New flow: use APPFND_CONHOS_UMS_URL env var directly.
+        #    Legacy flow: use the URL from the destination itself.
+        ums_url_override = os.environ.get(ENV_UMS_URL)
+        if ums_url_override:
+            base_url = ums_url_override
+            logger.debug(
+                "Using UMS URL from %s: %s", ENV_UMS_URL, base_url
+            )
+        else:
+            base_url = dest.url
```
NicoleMGomes: URL will never be on destination

hariturangi: The SDK has other options for passing the destination from a config parameter or through an environment variable. These options existed before; I assume they are provided for non-appfnd use cases/local testing or other purposes.

NicoleMGomes: For local testing the flow would be to export the variable still, no need to have it only for testing. 

Usually sdk offers a explicit object for configuration and the automatic resolution. Ideally we would keep this pattern.

hariturangi: removed the code that reads the url from destination

NicoleMGomes (dismissed):

NicoleMGomes (dismissed):

NicoleMGomes (approved):

PR #287, opened 2026-08-26
fix(telemetry): suppress high-cardinality httpx spans for MCP calls

jeanscherf:
## Summary

- Agents making many outbound calls produce traces with tens of thousands of spans because `HTTPXClientInstrumentor` creates a child span for every httpx request
- `excluded_urls` cannot be used as a fix: it skips **both** span creation and W3C `traceparent` header injection, which breaks distributed trace context propagation to downstream services
- Instead, a sync/async `request_hook` marks spans for URLs matching `SAP_CLOUD_SDK_HIGH_CARDINALITY_URLS` with `sap.cloud_sdk.suppress=True`, and `GenAIAttributeTransformer` filters them at export time — the span is created (so traceparent propagates correctly) but never exported to the collector
- Suppression is **opt-in via env var** (no default patterns) — all spans are exported by default so call visibility is preserved

## Behaviour

| | Default (unchanged) | With env var set |
|---|---|---|
| Spans exported per matching call | 1 span | 0 spans |
| W3C traceparent header injected | yes | yes |

Configure via env: `SAP_CLOUD_SDK_HIGH_CARDINALITY_URLS=pattern1,pattern2`

## Test plan

- [ ] `tests/core/unit/telemetry/test_genai_attribute_transformer.py` — two new tests: spans marked `sap.cloud_sdk.suppress=True` are filtered; suppressed-only batches pass an empty list to the wrapped exporter
- [ ] All 406 telemetry unit tests pass


jeanscherf:
Closing: the httpx instrumentation is working correctly. One span per outbound request is standard OTel behavior. Monster traces from high fan-out agents are a signal about agent design, not an instrumentation bug — suppressing those spans in the SDK would hide useful diagnostic information. The a2a-python-sdk EventQueue polling pattern is a separate issue to be addressed upstream.

PR #290, opened 2026-08-27
feat(destination, agent-memory, core): support binding rotation in HTTP clients

cassiofariasmachado:

## Description

Introduces a shared `HttpClient` + `XsuaaAuthProvider` layer in `core/` that all service module transports compose. Solves a silent failure mode where long-lived clients hold frozen credentials — after a BTP secret rotation the old token expires and every call fails with 401 until the client is re-instantiated.

Two layers of rotation resilience:
- **Proactive**: `ConfigFactory.has_changed()` checks the binding directory mtime before serving each cached token; on change it evicts the full token cache and re-reads credentials immediately.
- **Reactive**: `HttpClient.request` invalidates the token cache entry and retries once on any 401 response.

Migrates both `destination` and `agent_memory` modules to compose `HttpClient` directly (removes `HttpTransport` from `agent_memory`; replaces `TokenProvider` in `destination`). Rewrites unit test suites for both modules to use module-level `mock_http` fixtures and a `_make_response` helper.

## Related Issue

No linked issue.

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Dependency update

## How to Test

1. Run `uv run pytest tests/agent_memory/` — 248 passed, 12 skipped
2. Run `uv run pytest tests/destination/` — all passing
3. Run `uvx pre-commit run --all-files` — all hooks pass (ruff-format, ruff-check, ty)

## Checklist

- [x] All tests pass locally

## Breaking Changes

None. `AgentMemoryClient` and `DestinationClient` public APIs are unchanged. Internal `HttpTransport` class is removed but was never part of the public API.

## Additional Notes

- `auditlog` and `data_anonymization` modules are out of scope for this PR and can adopt `HttpClient` in a follow-up.
- `data_anonymization` uses mTLS and would need an `MTlsAuthProvider` implementation.
- `ConfigFactory` is generic (`ConfigFactory[C]`) and can be reused by any future module that needs binding re-read on rotation.


src/sap_cloud_sdk/agent_memory/_http.py
```diff
@@ -0,0 +1,71 @@
...
+from typing import Any, Optional
+from urllib.parse import quote, urlencode
+
+from requests.exceptions import RequestException, Timeout
+
+from sap_cloud_sdk.agent_memory.exceptions import (
+    AgentMemoryHttpError,
+    AgentMemoryNotFoundError,
+)
+from sap_cloud_sdk.core._http_client import HttpClient, HttpMethod
```
NicoleMGomes: Do we really to encapsulate it on a specific request method for agent memory? It seems to me this is being used more for exceptions, but not sure if this is really required.

src/sap_cloud_sdk/agent_memory/config.py
```diff
@@ -163,24 +166,43 @@ def _load_config_for_instance(instance: str) -> AgentMemoryConfig:
...
+            return binding.extract_config()
+        except AgentMemoryConfigError:
+            raise
+        except Exception as exc:
+            raise AgentMemoryConfigError(
+                f"Failed to load Agent Memory configuration for instance '{instance}': {exc}"
+            ) from exc
+
+    return ConfigFactory(
+        module="hana-agent-memory",
```
NicoleMGomes: I believe we already have some enum for telemetry, can we extract it to core and reuse?

src/sap_cloud_sdk/agent_memory/config.py
```diff
@@ -163,24 +166,43 @@ def _load_config_for_instance(instance: str) -> AgentMemoryConfig:
...
+        return _make_agent_memory_factory(instance)()
     except AgentMemoryConfigError:
         raise
     except Exception as exc:
         raise AgentMemoryConfigError(
             f"Failed to load Agent Memory configuration for instance '{instance}': {exc}"
         ) from exc
+
+
+def _make_agent_memory_factory(instance: str) -> "ConfigFactory[AgentMemoryConfig]":
```
NicoleMGomes:
```suggestion
def _make_actory(instance: str) -> "ConfigFactory[AgentMemoryConfig]":
```
No need to add agent_memory if you are in the agent memory folder, context is already clear.

src/sap_cloud_sdk/destination/__init__.py
```diff
@@ -74,6 +77,20 @@
...
 
 
+def _build_destination_http(
+    instance: Optional[str], config: Optional[DestinationConfig]
+) -> HttpClient:
+    if config is not None:
+        auth = XsuaaAuthProvider(lambda: config)
+        binding = config
+    else:
+        factory = _make_destination_factory(instance or "default")
```
NicoleMGomes: I belive it would be better for us to have an constant for default instance in the destination module.

NicoleMGomes:
Great solution!

src/sap_cloud_sdk/destination/_http.py
```diff
@@ -1,311 +1,53 @@
...
-            extra_headers=headers,
+def _request(
+    http: HttpClient,
+    method: HttpMethod,
+    path: str,
+    *,
+    params=None,
+    json=None,
+    headers=None,
+    tenant_subdomain=None,
```
jeanscherf: Should we type these like we do in the agent memory one?
Nothing bad. Just consistency.

src/sap_cloud_sdk/agent_memory/_http.py
cassiofariasmachado: Agree, maybe we could move it to the client as a class method. What do you think?

src/sap_cloud_sdk/core/_http_client.py
```diff
@@ -0,0 +1,242 @@
...
+            path: Path appended to ``base_url``. Should start with ``/``.
+            tenant_subdomain: Subscriber tenant subdomain forwarded to the auth
+                provider for per-tenant token derivation.
+            **kwargs: Forwarded verbatim to :meth:`requests.Session.request`.
+
+        Returns:
+            Raw :class:`requests.Response`. Callers must check the status code.
+        """
+        response = self._execute(method, path, tenant_subdomain, **kwargs)
+        if response.status_code == 401 and self._auth_provider is not None:
```
jeanscherf: I'd like to see a test with this path (did I miss it?):
The 401 → invalidate → retry path in HttpClient.request

src/sap_cloud_sdk/core/secret_resolver/_config_factory.py
```diff
@@ -0,0 +1,80 @@
...
+            base_volume_mount=self._base_volume_mount,
+            base_var_name=self._base_var_name,
+            module=self._module,
+            instance=self._instance,
+            target=binding,
+        )
+        binding.validate()
+        return self._extract(binding)
+
+    def has_changed(self) -> bool:
```
jeanscherf, line 73: Also I'd like a test with this branch:
consider the baseline, the mtime change detection and the OSError guard.

src/sap_cloud_sdk/core/_http_client.py
```diff
@@ -0,0 +1,242 @@
...
+            raise RuntimeError(f"Failed to obtain OAuth2 token: {exc}") from exc
+
+        expires_in: int = token.get("expires_in", 3600)
+        expires_at = datetime.now() + timedelta(
+            seconds=expires_in - _TOKEN_EXPIRY_BUFFER_SECONDS
+        )
+
+        existing = self._cache.get(tenant_subdomain)
+        if existing is not None:
+            existing[0].close()
```
tiagoek: Liked the overall structure — the `has_changed()` + 401 retry is exactly the pattern we converged on in PR #256 for AI Core.

One thing caught my attention here: if two concurrent requests arrive for the same tenant (common in agents), both can fall into `_fetch_token()`. The first closes the `OAuth2Session` while the second may still be using it for an in-flight request — that surfaces as a `ConnectionError`, not a `401`, so the retry in `HttpClient.request` won't catch it. What do you think about deferring the `close()`, or at least putting a lock around `_cache`?

src/sap_cloud_sdk/core/_http_client.py
```diff
@@ -0,0 +1,242 @@
...
+            tenant_subdomain: Subscriber tenant subdomain forwarded to the auth
+                provider for per-tenant token derivation.
+            **kwargs: Forwarded verbatim to :meth:`requests.Session.request`.
+
+        Returns:
+            Raw :class:`requests.Response`. Callers must check the status code.
+        """
+        response = self._execute(method, path, tenant_subdomain, **kwargs)
+        if response.status_code == 401 and self._auth_provider is not None:
+            self._auth_provider.invalidate(tenant_subdomain)
```
tiagoek: On the reactive retry, `invalidate(tenant_subdomain)` evicts only that tenant's token. Since a binding rotation changes the global `client_secret`, all tenants will eventually hit a 401 — but each at their own token expiry time, so the current per-tenant retry handles it correctly. Not a bug, just worth noting: if you wanted to be more aggressive and avoid any 401 for the remaining tenants after the first one fires, calling `invalidate_all()` here would flush them all at once. The `has_changed()` path already covers this proactively anyway.

src/sap_cloud_sdk/agent_memory/_http.py
cassiofariasmachado: I did this change, please tell me if you though something different. The `_request` is now a private method on `AgentMemoryClient`. It uses `self._http` directly and keeps the same error-mapping and OData encoding logic.

src/sap_cloud_sdk/agent_memory/config.py
cassiofariasmachado: Done. Renamed to `_make_config_factory`.

src/sap_cloud_sdk/destination/_http.py
cassiofariasmachado: Done.

src/sap_cloud_sdk/core/_http_client.py
cassiofariasmachado: No, I missed it heheh

Now I think the test suite is more complete. The following scenarios were added:

- `test_401_triggers_invalidate_and_retry`: 401 triggers invalidate() and a second request returns 200.
- `test_401_without_auth_provider_is_not_retried`: no retry when there's no auth provider.
- `test_200_response_not_retried`: happy path, no retry or invalidation.
- `test_retry_uses_fresh_session_after_invalidate`: second `get_session()` call returns a fresh session after invalidation.

src/sap_cloud_sdk/core/secret_resolver/_config_factory.py
cassiofariasmachado: Done.

src/sap_cloud_sdk/core/_http_client.py
cassiofariasmachado: Great catch. Added a lock around the `_cache`.

src/sap_cloud_sdk/core/_http_client.py
cassiofariasmachado: Good note. I am keeping `invalidate(tenant_subdomain)` as-is for the reactive path. The reasoning: a 401 may be tenant-specific (e.g. a subscriber token revoked for compliance reasons), so flushing all tenants on a single 401 would be wrong in these cases.

src/sap_cloud_sdk/agent_memory/client.py
```diff
@@ -102,18 +105,66 @@ def __init__(
...
+                tenant_subdomain=tenant_subdomain,
+                headers={"Content-Type": "application/json"},
+                **kwargs,
+            )
+        except Timeout as exc:
+            raise AgentMemoryHttpError(
+                f"Request timed out: {method.value} {path}"
+            ) from exc
+        except RequestException as exc:
+            raise AgentMemoryHttpError(
```
tiagoek: **[FLAG] PY-CON-01**

String literal 'agentID' appears 3× — extract module-level constant

src/sap_cloud_sdk/agent_memory/client.py
```diff
@@ -102,18 +105,66 @@ def __init__(
...
+                f"Request failed: {method.value} {path} — {exc}"
+            ) from exc
```
tiagoek: **[FLAG] PY-CON-01**

String literal 'content' appears 5× — extract module-level constant

tiagoek:
## SDK Module Review

| Check | Status | Findings |
|-------|--------|----------|
| bdd | ✅ PASS | 0 |
| binding-shape | ✅ PASS | 0 |
| commits | ✅ PASS | 0 |
| concurrency | ✅ PASS | 0 |
| constants | ⚠️ FLAG | 3 |
| deletion-hygiene | ✅ PASS | 0 |
| deps-supply | ✅ PASS | 0 |
| disclosure | ✅ PASS | 0 |
| docs | ✅ PASS | 0 |
| errors-logging | ✅ PASS | 0 |
| hardcode | ✅ PASS | 0 |
| http-hygiene | ⚠️ FLAG | 1 |
| license-spdx | ✅ PASS | 0 |
| patterns | ⚠️ FLAG | 1 |
| pr-size | ✅ PASS | 0 |
| quality-gate-parity | ✅ PASS | 0 |
| secrets | ✅ PASS | 0 |
| testing-depth | ✅ PASS | 0 |
| versioning | ✅ PASS | 0 |


### Findings (5)

- **[FLAG] PY-CON-01** — String literal 'agentID' appears 3× — extract module-level constant ([inline on `src/sap_cloud_sdk/agent_memory/client.py:145`](https://github.com/SAP/cloud-sdk-python/pull/290#discussion_r3883514547))
- **[FLAG] PY-CON-01** — String literal 'invokerID' appears 3× — extract module-level constant ([inline on `src/sap_cloud_sdk/agent_memory/client.py:146`](https://github.com/SAP/cloud-sdk-python/pull/290#discussion_r3883514670))
- **[FLAG] PY-CON-01** — String literal 'content' appears 5× — extract module-level constant ([inline on `src/sap_cloud_sdk/agent_memory/client.py:147`](https://github.com/SAP/cloud-sdk-python/pull/290#discussion_r3883514775))
- **[FLAG] HTTP-01** — HTTP session created per invocation — prefer single instance in __init__ ([inline on `src/sap_cloud_sdk/core/_http_client.py:250`](https://github.com/SAP/cloud-sdk-python/pull/290#discussion_r3883514925))
- **[FLAG] PY-PT-08** — Public function `create_client` missing type annotations: return type ([inline on `src/sap_cloud_sdk/destination/__init__.py:82`](https://github.com/SAP/cloud-sdk-python/pull/290#discussion_r3883515036))

_5 finding(s): 5 posted as inline comment(s) on the affected lines, 0 not tied to a code line (listed above)._

---
_Generated by sdk-review-skill · v1_

src/sap_cloud_sdk/agent_memory/client.py
tiagoek, line 164: **[FLAG] PY-CON-01**

String literal 'agentID' appears 3× — extract module-level constant

src/sap_cloud_sdk/agent_memory/client.py
```diff
@@ -102,18 +105,66 @@ def __init__(
...
+                headers={"Content-Type": "application/json"},
+                **kwargs,
+            )
+        except Timeout as exc:
+            raise AgentMemoryHttpError(
+                f"Request timed out: {method.value} {path}"
+            ) from exc
+        except RequestException as exc:
+            raise AgentMemoryHttpError(
+                f"Request failed: {method.value} {path} — {exc}"
```
tiagoek, line 165: **[FLAG] PY-CON-01**

String literal 'invokerID' appears 3× — extract module-level constant

src/sap_cloud_sdk/agent_memory/client.py
tiagoek, line 166: **[FLAG] PY-CON-01**

String literal 'content' appears 5× — extract module-level constant

src/sap_cloud_sdk/core/_http_client.py
```diff
@@ -0,0 +1,266 @@
...
+        if self._auth_provider is not None:
+            session: requests.Session = self._auth_provider.get_session(
+                tenant_subdomain
+            )
+            # Use the auth provider's current base_url (updated on every token
+            # refresh) so requests go to the correct URL after secret rotation.
+            base_url = getattr(self._auth_provider, "base_url", None) or self._base_url
+        else:
+            if self._plain_session is None:
+                self._plain_session = requests.Session()
```
tiagoek: **[FLAG] HTTP-01**

HTTP session created per invocation — prefer single instance in __init__

src/sap_cloud_sdk/destination/__init__.py
```diff
@@ -73,6 +76,22 @@
 
 logger = logging.getLogger(__name__)
 
+_DEFAULT_INSTANCE = "default"
+
+
+def _build_destination_http(
```
tiagoek, line 83: **[FLAG] PY-PT-08**

Public function `create_client` missing type annotations: return type

src/sap_cloud_sdk/agent_memory/config.py
cassiofariasmachado: I will let this refactor for later because it will impact all modules.

src/sap_cloud_sdk/core/protocol/http/models.py
```diff
@@ -0,0 +1,274 @@
...
+        self._config = config_factory()
+        self._timeout = timeout
+        self._cache: dict[Optional[str], tuple[OAuth2Session, datetime]] = {}
+        self._lock = threading.Lock()
+
+    def get_session(self, tenant_subdomain: Optional[str] = None) -> OAuth2Session:
+        has_changed = getattr(self._config_factory, "has_changed", None)
+        if callable(has_changed) and has_changed():
+            with self._lock:
+                logger.info("Invalidating token cache due to secret rotation")
```
cassiofariasmachado: Not sure if we keep these logs or not. Any thoughts about it?

src/sap_cloud_sdk/destination/config.py
```diff
@@ -37,6 +37,9 @@
 from sap_cloud_sdk.destination.exceptions import ConfigError
 from sap_cloud_sdk.destination._models import TransparentProxy
 
+if TYPE_CHECKING:
+    from sap_cloud_sdk.core.secret_resolver import ConfigFactory
+
 _TRANSPARENT_PROXY_ENV_VAR = "APPFND_CONHOS_TRANSP_PROXY"
```
betinacosta, line 43: Should this be duplicated?

src/sap_cloud_sdk/core/_http_client.py
```diff
@@ -0,0 +1,274 @@
+"""Shared HTTP client with injectable authentication and rotation-resilient token management.
```
NicoleMGomes: shouldn't this be inside a http or protocol folder / package?

betinacosta:
General comment: Shouldn't we update the user-documentations?

src/sap_cloud_sdk/core/_http_client.py
cassiofariasmachado: Done

src/sap_cloud_sdk/destination/config.py
cassiofariasmachado: Fixed.

src/sap_cloud_sdk/core/protocol/http/models.py
NicoleMGomes: I believe we can keep, but as debug

NicoleMGomes (dismissed):

NicoleMGomes (dismissed):

NicoleMGomes:
WhatsApp Image 2026-08-12 at 16 28 59

NicoleMGomes (dismissed):

NicoleMGomes (approved):

ArthurTonial (approved):
Image

PR #302, opened 2026-09-03
fix: pin versions

ArthurTonial:

## Description

Replaces `~=` (compatible release) pin specifiers with explicit `>=x,<N` major-version ranges for all runtime dependencies in `pyproject.toml`. This makes upper bounds intentional and documented rather than an implicit side-effect of the `~=` operator.

Also bumps the `protobuf` lower bound from `>=4.25.0` to `>=5.0.0`.

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Dependency update

## How to Test

1. Run `uv sync` from a clean environment — resolution should succeed with no conflicts.
2. Run the full test suite: `uv run pytest`
3. Verify all dependency versions in `uv.lock` fall within the new declared ranges.

## Checklist

- [x] All tests pass locally

## Breaking Changes

None.


pyproject.toml
```diff
@@ -9,45 +9,30 @@ authors = [
...
 dependencies = [
-    "minio~=7.2.16",
-    "setuptools>=83.0",
-    "requests>=2.33.0",
-    "requests-oauthlib~=2.0.0",
-    "pydantic~=2.12.3",
-    "hatchling~=1.27.0",
-    "opentelemetry-exporter-otlp-proto-grpc~=1.43.0",
-    "opentelemetry-exporter-otlp-proto-http~=1.43.0",
-    "traceloop-sdk~=0.61.0",
```
LucasAlvesSoares, line 20: ˜= means compatible version so I believe those don't need to change

NicoleMGomes: Good point, we can keep ~=

ArthurTonial: Fixed

cassiofariasmachado (approved):

PR #320, opened 2026-09-09
fix(telemetry): merge SDK resource attrs into pre-installed MeterProvider

jeanscherf:
## Problem

Reported: `sap.internal_sdk.version` and `sap.cloud_sdk.name` were missing from **metrics** when OTel auto-instrumentation was active, even though they appeared correctly on traces and logs.

Root cause: when the OTel auto-instrumentation wrapper installs a `MeterProvider` before `auto_instrument()` is called, `metrics.set_meter_provider()` is a silent no-op (`WARNING: Overriding of current MeterProvider is not allowed`). The new provider with our resource attributes is created but never registered, so the global provider (owned by the wrapper) never gets our attributes.

Traces and logs already had a fix for this — they detect the pre-installed provider and merge resource attributes into it. Metrics had no equivalent.

## Fix

Mirror the log/trace pattern: at `_setup_meter_provider()` time, check if an SDK `MeterProvider` is already set globally. If so, merge our resource attributes into it instead of trying to create a new one.

`SdkConfiguration` (which stores the resource) is a dataclass, so `_sdk_config.resource` is mutable in-place. `_measurement_consumer` holds a reference to the same object, so it sees the update automatically — no additional patching needed.

## Tests

- `TestMergeSdkResourceIntoMeterProvider` — unit tests for the merge function itself (resource update, measurement consumer sees it, SDK attrs win on collision)
- `TestSetupMeterProvider.test_reuses_existing_sdk_meter_provider` — verifies merge path is taken and `set_meter_provider` is not called again
- `TestSetupMeterProvider.test_existing_provider_no_new_reader` — verifies no extra `PeriodicExportingMetricReader` is created on the reuse path

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "sap-cloud-sdk"
-version = "0.52.0"
+version = "0.52.1"
 description = "SAP Cloud SDK for Python"
 readme = "README.md"
 license = "Apache-2.0"
--- a/src/sap_cloud_sdk/core/telemetry/_provider.py
+++ b/src/sap_cloud_sdk/core/telemetry/_provider.py
@@ -30,14 +30,14 @@ from opentelemetry.sdk.metrics import (
     ObservableUpDownCounter,
     UpDownCounter,
 )
-
-# Stable reference for isinstance checks — not overwritten when tests patch MeterProvider
-_SDKMeterProvider = MeterProvider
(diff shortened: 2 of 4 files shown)


NicoleMGomes (dismissed):

ArthurTonial (approved):

ArthurTonial (approved):
Image

PR #322, opened 2026-09-09
perf(agentgateway): parallelize LoB MCP tool listing across fragments

### LINKED ISSUE #301: Parallelize fragment connections inside list_mcp_tools so all fragments are handled concurrently
### Describe the Problem

Saw some changes being made for `list_mcp_tools`:
- https://github.com/SAP/cloud-sdk-python/releases/tag/v0.41.0
- https://github.com/SAP/cloud-sdk-python/issues/178

From my understanding `list_mcp_tools()` fetches the registered MCP fragments sequentially.
That means fetch time scales with the number of such fragments.
> N fragments x ~3s ( equals ~44 seconds for N=14 for example)

### Propose a Solution

Parallelizing these fragment fetches could further improve performance, especially when there is no cache yet.
Fetch time would not scale with the number of MCP fragments.

### Describe Alternatives

_No response_

### Affected Development Phase

Getting Started

### Impact

No Impact

### Timeline

_No response_

wangzhengzhuo05:
Fixes #301

## What / Why

`get_mcp_tools_lob` fetched MCP tools from destination fragments **sequentially** — N fragments × ~3s network round-trip each. Concurrent fetching makes total time ~max(fragment time) instead of the sum (e.g. 14 fragments ≈ 3s instead of ≈44s).

## How

- Replaced the sequential `for fragment ... await list_server_tools` loop with `asyncio.gather(*coros, return_exceptions=True)`.
- URL-missing fragments are still skipped (with warning) at scheduling time; failing fragments are isolated via `return_exceptions=True` + `_log_mcp_server_error` — same continuation semantics as before.
- Results are zipped back in fragment order (gather returns in input order), so tool ordering is unchanged.
- Pre-fetch ORD filter, post-fetch names filter, and logging are untouched.

## Tests

- `test_lists_fragments_concurrently`: 3 fragments × 0.05s sleeps complete in < 2× per-call (sequential would be 3×). **Mutation check**: reverting to a sequential loop makes this test fail (verified).
- `test_failure_in_one_fragment_does_not_fail_others`: 2nd fragment raises; other fragments' tools still returned.
- `test_preserves_fragment_order`: completion order is inverted via varied latencies; returned tools still follow fragment order.

Local: 168 passed (test_lob + test_agw_client + test_customer).

### AI-generated code disclosure

This change was implemented with the assistance of AI (OpenCode + muse-spark) and reviewed by a human maintainer process; per SAP GenAI guidelines the code has been verified for correctness, security, and license compatibility.

### DIFF
--- a/src/sap_cloud_sdk/agentgateway/_lob.py
+++ b/src/sap_cloud_sdk/agentgateway/_lob.py
@@ -450,6 +450,7 @@ async def get_mcp_tools_lob(
             in ord_ids_set
         ]
 
+    pending: list[tuple[str, str]] = []
     for fragment in fragments:
         fragment_name = fragment.name
         mcp_url = fragment.properties.get("URL") or fragment.properties.get("url")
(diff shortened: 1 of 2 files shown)


wangzhengzhuo05:
Closing this to keep the tracker clean \u2014 it has been sitting without review for several days. This is no reflection on the change itself: the fix and its local verification still stand. If the issue is still present and you would like the patch revisited, say the word and I will rebase it onto current main and re-open. Thanks for your time.

PR #327, opened 2026-09-11
fix(agentgateway): raise AgentGatewayServerError on isError tool results

### LINKED ISSUE #326: AgentGatewayClient.call_mcp_tool returns None instead of CallToolResult when MCP server responds with isError: true
### Describe the Bug

When AgentGatewayClient.call_mcp_tool is used to invoke an MCP tool and the MCP server returns a spec-compliant error response (isError: true, HTTP 200, SSE event: message), the method returns None instead of a CallToolResult. Any subsequent access on the return value raises AttributeError: 'NoneType' object has no attribute 'isError'.

The MCP server side is correct — a direct tools/call against the MCP endpoint returns a well-formed response (see Expected Behavior). The failure is in how call_mcp_tool parses the SSE response for the isError: true case.

### Steps to Reproduce

Set up an AgentGatewayClient pointing to an MCP server deployed via Agent Gateway.
Discover tools via agw_client.list_mcp_tools(user_token=...) and locate a tool whose backend can return an error (e.g. a lookup tool called with an ID that does not exist).
Call await agw_client.call_mcp_tool(tool, user_token=..., <arg>=<value-that-triggers-backend-error>).
The MCP server returns HTTP 200 + SSE event: message with isError: true (confirmed via direct call — see Expected Behavior).
call_mcp_tool returns None.
Access result.isError → AttributeError: 'NoneType' object has no attribute 'isError'.

### Expected Behavior

A direct tools/call to the same MCP server with the same arguments returns a well-formed, spec-compliant response (MCP protocol 2025-11-25):

HTTP 200 — text/event-stream

event: message
data: {
  "result": {
    "content": [{"type": "text", "text": "{\"status\":500,\"error\":\"HTTP Error\",\"message\":\"Tool call failed\"}"}],
    "isError": true
  },
  "jsonrpc": "2.0",
  "id": 3
}
call_mcp_tool should parse this SSE event and return a CallToolResult with isError=True and content populated. If parsing fails for any reason, it should raise an exception — returning None silently is never a valid outcome.

### Screenshots

_No response_

### Used Versions

sap-cloud-sdk          0.48.2
fastapi                0.135.3
langchain-core         1.5.3
langchain-mcp-adapters (version — run uv pip list to confirm)

### Code Examples

```python
# Your code here
```

### Stack Trace

AttributeError: 'NoneType' object has no attribute 'isError'

The above exception was the direct cause of the following exception:

  File "app/clients/mcp_validation_rule_wrapper.py", line 197, in _call_tool
    result = await _invoke_with_timeout(tool, arguments)
  # (exception originates inside sap_cloud_sdk.agentgateway internals
  #  when accessing result.isError on the None returned by call_mcp_tool)

### Log File

INFO  [mcp_wrapper] call tool=<tool-name> args=[...]
INFO  Negotiated protocol version: 2025-11-25
INFO  Loaded 7 MCP tool(s) from Agent Gateway (deployed mode)
ERROR Unexpected error during tool invocation
ERROR [mcp_wrapper] tool=<tool-name> error=Tool invocation failed for '<tool-name>':
      'NoneType' object has no attribute 'isError'
INFO  [mcp_wrapper] done tool=<tool-name> duration_ms=9603 status=error

### Affected Development Phase

Development

### Impact

Impaired

### Timeline

_No response_

wangzhengzhuo05:
## What

`AgentGatewayClient.call_mcp_tool` silently swallowed MCP tool errors. When the MCP server returns a spec-compliant error response (`isError: true`, HTTP 200, SSE `event: message` with `result.isError == true`), both `call_mcp_tool_lob` (LoB flow) and `call_mcp_tool_customer` (customer flow) logged the error and returned the raw error text as if it were a successful tool result. Callers could not distinguish success from failure — any caller inspecting `result.isError` on the return value crashed with `AttributeError` (reported as `'NoneType' object has no attribute 'isError'`).

## Why

The SDK already documents the intended contract in `src/sap_cloud_sdk/agentgateway/exceptions.py`: `AgentGatewayServerError` is raised when "A tool invocation returns an error result (isError=True)". The implementation violated its own documented contract — the exception was never raised on this path. Silently returning a value that crashes `.isError` accessors is never a valid outcome for a tool call that failed server-side.

## How

- `src/sap_cloud_sdk/agentgateway/_lob.py`: in `call_mcp_tool_lob`, replace the log-and-return block with `raise AgentGatewayServerError(f"Tool '{tool.name}' on '{tool.url}' returned an error: {text}")` when `mcp_is_error(result)` is true.
- `src/sap_cloud_sdk/agentgateway/_customer.py`: identical fix in `call_mcp_tool_customer` so both flows behave consistently.
- Success path unchanged (still returns the tool text); empty-content path unchanged (still returns `""`).
- Tests: added `test_raises_server_error_on_is_error` to `TestCallMcpToolLob` and `TestCallMcpToolCustomer`; set explicit `is_error=False`/`isError=False` on the existing success-path mocks (a plain `MagicMock` is truthy, so those tests were implicitly passing through the error branch before).

## Verification

- Reproduced the bug at the wire level: a local HTTP server answering the exact SSE payload from the issue (`event: message` / `data: {"result": {"content": [...], "isError": true}, "jsonrpc": "2.0", "id": 3}`) against `call_mcp_tool_lob` returned the error text as a plain `str` and `.isError` raised `AttributeError: 'str' object has no attribute 'isError'`.
- Targeted suite: `pytest tests/agentgateway/unit/test_lob.py tests/agentgateway/unit/test_customer.py` → 103 passed.
- Full agentgateway unit suite: 220 passed, 24 failed — the 24 failures are in `test_converters.py` and **pre-exist on a clean `origin/main` checkout** (verified via a fresh worktree: same 24 failures, unrelated to this change).
- Mutation check: reverted only the fix's behavior in `_lob.py` (restored log-and-return) → `test_raises_server_error_on_is_error` fails (1 failed); restored the fix → passes. The new test guards this behavior.

## Notes

- The reporter's exact stack (`'NoneType' object has no attribute 'isError'`) comes from an older SDK version; on current main the returned value is a `str`, producing `'str' object has no attribute 'isError'` — same class of defect, same fix.
- No new dependencies; no formatting churn.

Fixes #326

### AI-generated code disclosure

This change was implemented with AI assistance (OpenCode + muse-spark) and reviewed/verified for correctness, security, and license compatibility per SAP GenAI guidelines.

### DIFF
--- a/src/sap_cloud_sdk/agentgateway/_customer.py
+++ b/src/sap_cloud_sdk/agentgateway/_customer.py
@@ -46,7 +46,10 @@ from sap_cloud_sdk.agentgateway._models import (
     MCPToolFilter,
 )
 from sap_cloud_sdk.agentgateway._token_cache import _TokenCache
-from sap_cloud_sdk.agentgateway.exceptions import AgentGatewaySDKError
+from sap_cloud_sdk.agentgateway.exceptions import (
+    AgentGatewaySDKError,
+    AgentGatewayServerError,
+)
 from sap_cloud_sdk.core.secret_resolver import resolve_base_mount
 
 logger = logging.getLogger(__name__)
@@ -841,11 +844,8 @@ async def call_mcp_tool_customer(
                 text = str(getattr(first, "text", ""))
 
                 if mcp_is_error(result):
-                    logger.error(
-                        "Tool '%s' on '%s' returned an error: %s",
-                        tool.name,
-                        tool.url,
-                        text,
+                    raise AgentGatewayServerError(
+                        f"Tool '{tool.name}' on '{tool.url}' returned an error: {text}"
                     )
 
                 return text
--- a/src/sap_cloud_sdk/agentgateway/_lob.py
+++ b/src/sap_cloud_sdk/agentgateway/_lob.py
@@ -50,6 +50,7 @@ from sap_cloud_sdk.agentgateway._models import (
 from sap_cloud_sdk.agentgateway._token_cache import _GatewayUrlCache, _TokenCache
 from sap_cloud_sdk.agentgateway.exceptions import (
     AgentGatewaySDKError,
+    AgentGatewayServerError,
     MCPServerNotFoundError,
 )
 
@@ -526,11 +527,8 @@ async def call_mcp_tool_lob(
                 text = str(getattr(first, "text", ""))
 
                 if mcp_is_error(result):
-                    logger.error(
-                        "Tool '%s' on '%s' returned an error: %s",
-                        tool.name,
-                        tool.url,
-                        text,
+                    raise AgentGatewayServerError(
+                        f"Tool '{tool.name}' on '{tool.url}' returned an error: {text}"
                     )
 
                 return text
--- a/tests/agentgateway/unit/test_customer.py
+++ b/tests/agentgateway/unit/test_customer.py
@@ -27,7 +27,10 @@ from sap_cloud_sdk.agentgateway._models import (
 )
 from sap_cloud_sdk.agentgateway._token_cache import _TokenCache
 from sap_cloud_sdk.agentgateway.config import ClientConfig
-from sap_cloud_sdk.agentgateway.exceptions import AgentGatewaySDKError
+from sap_cloud_sdk.agentgateway.exceptions import (
+    AgentGatewaySDKError,
+    AgentGatewayServerError,
+)
 
 
 # ============================================================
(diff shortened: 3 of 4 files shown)


wangzhengzhuo05:
Closing this to keep the tracker clean \u2014 it has been sitting without review for several days. This is no reflection on the change itself: the fix and its local verification still stand. If the issue is still present and you would like the patch revisited, say the word and I will rebase it onto current main and re-open. Thanks for your time.

PR #329, opened 2026-09-11
docs(auditlog_ng): update user-guide

jplbrun:
> **Disclaimer:** Do not include SAP-internal or customer-specific information in this PR (e.g. internal system URLs, customer names, tenant IDs, or confidential configurations). This is a public repository.

## Description

Reworks the `auditlog_ng` user guide (`src/sap_cloud_sdk/core/auditlog_ng/user-guide.md`) so it follows the conventions used by the other module user guides, and corrects several statements that did not match the actual client behavior.

**Structure aligned to the house convention**

- Retitled to `# Audit Log NG User Guide` and merged the redundant `Overview` into the single intro paragraph.
- Renamed `Prerequisites` → `Installation`, added a `Quick Start` section (minimal `create_client(tenant=...)` → build event → `send()` → `close()`), and grouped the walkthrough, full example, one-off sends, serialization table, and auto-injection under a single `Usage` section.
- Normalized headings to Title Case, removed the ✅/❌ emoji from the config tables in favor of plain `Yes`/`No`, and dropped the duplicate import block and the module-specific `Running the Unit Tests` section.

**Corrected content (verified against the code)**

- `Validation` section replaced with an `Error Handling` section that lists the exceptions `send()` actually raises: `ValidationError` on schema failure (subclass of `AuditLogNGError`), `ValueError` for an invalid `common.tenant_id`, and `RuntimeError` on a closed client. The previous text incorrectly claimed a `tenant_id` UUID check and used the wrong exception types.
- Fixed the config-path description to reflect the real precedence (`config=` → `tenant` → explicit kwargs).
- Matched the validation regex shown in the docs (`[a-zA-Z0-9._/~-]`) to the pattern compiled in `config.py`.

**Recommended setup updated**

- Automatic tenant/user injection now documents `bootstrap(app)` as the recommended way to wire IAS context + telemetry for new apps, with `StarletteIASTelemetryMiddleware` retained under a `Legacy` subsection for apps not yet on `bootstrap()`.

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Open `src/sap_cloud_sdk/core/auditlog_ng/user-guide.md` and read it top to bottom

## Checklist

[...]

PR #330, opened 2026-09-11
fix(ci): make version-bump check against merge-base

jplbrun:

## Description

The **Check Version Bump** workflow diffs `github.event.pull_request.base.sha` against `head.sha` to decide whether a version bump is required. That base SHA is the commit `main` pointed at when the PR was last synced, and it drifts behind `main` as other PRs merge. As a result the diff includes `src/**/*.py` files from already-merged PRs that the current branch never touched, so the check wrongly demands a version bump and a documentation-only PR ends up failing because unrelated source files appear in the comparison.

This changes the check to diff against the **merge-base** of the base and head commits (the point where the branch actually diverged from `main`) instead of the raw base ref. 

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

Reproduced locally using the workflow's own logic against a drifted PR's real SHAs:

1. Old behavior (raw `base.sha`): source files from already-merged PRs appear in the diff → check **FAILS** demanding a version bump.
2. New behavior (merge-base): only the branch's own files appear → docs-only diff prints `No source file changes under src/. Version bump not required.` → check **PASSES**.
3. Negative case preserved: a diff that genuinely contains a `src/**/*.py` change without a version bump still matches the regex and **FAILS** with the existing error — enforcement is unchanged for real source changes.

Also tested end-to-end : applying the fix to docs-only PR branch (#329) flipped **Check Version Bump** from failing to passing.

## Checklist

- [x] All tests pass locally

## Breaking Changes

None

## Additional Notes

None

### DIFF
--- a/.github/workflows/check-version-bump.yaml
+++ b/.github/workflows/check-version-bump.yaml
@@ -22,11 +22,12 @@ jobs:
         run: |
           BASE_SHA="${{ github.event.pull_request.base.sha }}"
           HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+          MERGE_BASE=$(git merge-base "$BASE_SHA" "$HEAD_SHA")
 
-          CHANGED_FILES=$(git diff --name-only "$BASE_SHA" "$HEAD_SHA")
+          CHANGED_FILES=$(git diff --name-only "$MERGE_BASE" "$HEAD_SHA")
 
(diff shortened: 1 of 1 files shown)


jeanscherf (approved):
Great catch!

PR #332, opened 2026-09-14
feat: add retry credentials to all modules using binding

NicoleMGomes:

## Description

Adds binding rotation support to all SDK modules that read credentials from BTP service bindings (volume mounts or environment variables). When a BTP binding rotates at runtime, long-lived clients previously held stale credentials and would fail until the application was restarted. This change makes every module resilient to secret rotation without requiring a restart or client recreation.

Two resilience layers are applied consistently across all affected modules:

- **Proactive**: Before every credential use, `ConfigFactory.has_changed()` checks whether the secret directory mtime has changed. If so, credentials are re-read and any cached tokens or sessions are discarded.
- **Reactive** (object store only): On an `InvalidAccessKeyId` or `SignatureDoesNotMatch` S3 error, credentials are refreshed and the failed operation is retried once.

Modules updated: `objectstore`, `print`, `adms`, `dms`, `auditlog`.

All changes are backward-compatible. The `config=` explicit parameter path continues to work on all `create_client` functions; callers using explicit config simply do not get rotation tracking.

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [ ] Dependency update

## How to Test

1. Run the unit test suites for all affected modules:
   ```bash
   uv run pytest tests/objectstore/unit/ tests/print/unit/ tests/adms/unit/ tests/dms/unit/ tests/core/unit/auditlog/ -v
   ```
2. Confirm all rotation-specific test classes pass:
   - `tests/objectstore/unit/test_s3_client.py::TestRotation`
   - `tests/print/unit/test_http.py::TestTokenProviderRotation`
   - `tests/adms/unit/test_ias_fetcher.py::TestIasTokenFetcherRotation`
   - `tests/dms/unit/test_auth.py::TestAuthRotation`
   - `tests/core/unit/auditlog/unit/test_http_transport.py::TestHttpTransportRotation`
3. Run pre-commit checks: `uvx pre-commit run --all-files`
4. For end-to-end verification: deploy an app on BTP using one of the affected modules, trigger a binding rotation on the service instance, and confirm requests continue to succeed without a restart.

## Checklist

- [x] All tests pass locally

## Additional Notes

The rotation pattern is now documented in `docs/GUIDELINES.md` under "Credential Binding Rotation" and is enforced by the code review skill. Any future module that reads from service bindings should follow the same pattern: accept a `ConfigFactory` on the auto-detection path, implement `_refresh_if_rotated()`, and include the four standard rotation test cases (proactive, reactive, no-rebuild, static config).

Internal note: `ObjectStoreClient.__init__` now accepts a `ConfigFactory[ObjectStoreBindingData]` instead of a plain `ObjectStoreBindingData`. This is not a public API break since `ObjectStoreClient` is intended to be constructed via `create_client()`, but any code directly instantiating `ObjectStoreClient` will need to wrap the config in a factory callable.


src/sap_cloud_sdk/dms/client.py
```diff
@@ -98,14 +99,14 @@ def __init__(
             authentication and handles environment detection.
 
         Args:
-            credentials: OAuth2 credentials and service URI for the DMS instance.
+            credentials: OAuth2 credentials (or a factory returning them) for the DMS instance.
             connect_timeout: TCP connection timeout in seconds. Defaults to 10.
             read_timeout: Response read timeout in seconds. Defaults to 30.
         """
         auth = Auth(credentials)
         self._http: HttpInvoker = HttpInvoker(
```
cassiofariasmachado, line 107: Can’t we replace this `HttpInvoker` with our implementation of the `HttpClient` from `sap_cloud_sdk.core.protocol.http`?

Please review the other modules.

cassiofariasmachado (changes requested):

src/sap_cloud_sdk/dms/client.py
NicoleMGomes: Good point, done.

src/sap_cloud_sdk/objectstore/_s3.py
```diff
@@ -1,8 +1,11 @@
 """S3 backend implementation for object store operations using MinIO client."""
 
 import io
+import logging
 import os
 import threading
+
+logger = logging.getLogger(__name__)
```
cassiofariasmachado: I think it's better to move this `logger` to after the imports.

src/sap_cloud_sdk/adms/_ias_fetcher.py
```diff
@@ -14,10 +14,13 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Callable, Optional
 
 import requests
 
+logger = logging.getLogger(__name__)
```
cassiofariasmachado: Move to after the imports.

cassiofariasmachado (dismissed):
Great work! I just added two minor suggestions.

src/sap_cloud_sdk/objectstore/_s3.py
NicoleMGomes: Done

src/sap_cloud_sdk/adms/_ias_fetcher.py
NicoleMGomes: Done

jeanscherf (approved):

ArthurTonial (approved):

PR #345, opened 2026-09-18
fix: validate destination tenant subdomain

betinacosta:

## Description

Adds strict RFC 1123 DNS-label validation to `XsuaaAuthProvider._fetch_token` before the subscriber tenant subdomain is interpolated into the OAuth token URL.

Previously, the `tenant` string accepted by every public Destination (and other service) client method was forwarded unchanged into a `str.replace` call on the token URL. Because `str.replace` has no URL awareness, a value containing URL delimiters — for example `evil.example/oauth/token?ignore=` — would change the parsed authority of the token URL to an attacker-chosen hostname while moving the trusted authentication suffix into a query parameter. The binding's `client_id` and `client_secret` would then be posted to that endpoint; standard TLS verification does not block this because it verifies the attacker-chosen authority rather than the intended one.

The fix calls `_validate_tenant_subdomain` before any URL manipulation. The function enforces the regex `^[A-Za-z0-9](?:[A-Za-z0-9\-]{0,61}[A-Za-z0-9])?$` and raises `ValueError` immediately if the value is not a single valid DNS label. This ensures only the provider identity-zone label in the token URL can be replaced, preserving the trusted authentication origin.

The validation is placed in `XsuaaAuthProvider` (`sap_cloud_sdk.core.protocol.http.models`) rather than in any individual service module so that all SDK clients using XSUAA client-credentials flow are protected uniformly.

## Related Issue

Closes #

## Type of Change

## How to Test

1. Install the package from this branch.
2. Construct an `XsuaaAuthProvider` with a synthetic config (any `token_url`, `identityzone`, `client_id`, `client_secret`).
3. Call `provider._fetch_token("evil.example/oauth/token?ignore=")` — expect `ValueError: Invalid tenant_subdomain`.
4. Call `provider._fetch_token("tenant-123")` — expect normal token acquisition (no error).
5. Run the targeted test class:

```bash
uv run pytest tests/core/unit/test_http_client.py::TestXsuaaAuthProviderSubdomainValidation -v
```

All 6 new tests should pass; no existing tests should regress.

## Checklist

- [x] All tests pass locally

## Additional Notes

- `_validate_tenant_subdomain` is intentionally private (`sap_cloud_sdk.core._tenant`). It is not part of the public API.
- Provider-context calls (where `tenant_subdomain` is `None`) are unaffected; the validation is guarded by `tenant_subdomain is not None`.
- Consumers that currently pass syntactically invalid tenant strings will now receive a `ValueError` at token-acquisition time rather than silently routing credentials to an unintended endpoint. This is the intended behavior change.


jeanscherf (approved):
Great work!

